PR #10, opened 2012-12-07
Fixes #3063 : add reporting to know if Apache2 was started or not

ncharles:


peckpeck:
@ncharles 
This PR is not mergeable anymore and is for and old branch.
The ticket is in discussion, I think we should close it

ncharles:
indeed, but could you suggest in the ticket a solution for the problem ?

jooooooon:
This has already been implemented in http://www.rudder-project.org/redmine/issues/3025.

PR #29, opened 2013-01-09
Fixes #3165: Add new version of the Technique to support differentiating...

jooooooon:
... repos by OS/OS-version/OS-version-SP


fanf:
I'm closing that PR, because 2.3 ? Really ? 
A new PR, perhaps based on that one (but seems simpler to just start over with the current state of things) will have to be done.

PR #34, opened 2013-01-11
Refs #3178 - Add a new version of PT 'RUg/YaST Package Manager Configuration' which can deal with GPG key of repositories with Rug. Good reporting is still missing

nperron:
Refs #3178 - Add a new version of PT 'RUg/YaST Package Manager Configuration' which can deal with GPG key of repositories with Rug. Good reporting is still for now


peckpeck:
OK, so I'm closing it

PR #62, opened 2013-03-18
Refs #3361: Initial autoFS related Techniques implementation

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/3361

To be merged in the 2.4 branch.


fanf:
This PR does not target still supported version of Rudder. It will have to be open again to at least Rudder 2.10, with the remarks implemented, if wanted.

PR #243, opened 2013-11-24
Ust 3647/dev/3664 add support for addresses

ncharles:


ncharles:
replaced by https://github.com/Normation/rudder-techniques/pull/717/commits

PR #366, opened 2014-05-19
Fixes #4868 - Add possibility to monitor a file except if absent

nperron:
Fixes #4868 - Add possibility to monitor a file except if absent


peckpeck:
Already fixed

PR #371, opened 2014-05-19
Fixes #4720 - Add ignore_interfaces.st with a minimum of interfaces to ignores (those added by virtualbox and hinder CFEngine)

nperron:
Fixes #4720 - Add ignore_interfaces.st with a minimum of interfaces to ignores (those added by virtualbox and hinder CFEngine)

cf http://www.rudder-project.org/redmine/issues/4720


peckpeck:
This is not how we want to do it.
We want to make it configurable

PR #393, opened 2014-06-06
Fixes #4959 -  Permit non existent file to be ignored

nperron:
Fixes #4959 -  Permit non existent file to be ignored

cf http://www.rudder-project.org/redmine/issues/4959

Just add new functionnality: Checkbox to know if we want to ignore non-existent files or not.


peckpeck:
Not up to date

PR #444, opened 2014-07-17
Fixes #4443 - Fix reporting for user without password on userManagementv...

nperron:
Fixes #4443 - Fix reporting for user without password on userManagementv2.0

cf http://www.rudder-project.org/redmine/issues/4443


peckpeck:
Ticket already rejected

PR #452, opened 2014-07-25
Fixes #5071 - Add new version of Technique in order to handle backports packages

nperron:
Fixes #5071 - Add new version of Technique in order to handle backports packages

cf http://www.rudder-project.org/redmine/issues/5071


amousset:
Closing as the target technique is deprecated.

PR #531, opened 2014-10-07
Fixes #5580: check the monotony of promises timestamp

ncharles:


peckpeck:
@ncharles This PR is not mergeable anymore
@jooooooon is your review positive ?

jooooooon:
On 3 April 2015 10:35:49 CEST, peckpeck <notifications@github.com> wrote:
>@ncharles This PR is not mergeable anymore
>@jooooooon is your review positive ?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/Normation/rudder-techniques/pull/531#issuecomment-89220484

If I recall correctly, we decided not to merge this change to reduce risk, and keep it for a better time. Maybe 3.1 is a good time for this? 

I don't think I have reviewed it, so don't worry. 
-- 
Sent from Kaiten Mail. Please excuse my brevity.

peckpeck:
@ncharles any update ?

jooooooon:
For what it's worth, I have now reviewed this code and it looks good to me. I just have one question: How does the rudder_tools_updated_origin file get created initially? The code here won't create it if it's missing, this is why I ask (but it could easily be changed to do so - maybe it should by default so that there is no risk of it not existing?)

I propose to merge this into master (for 3.2 / 4.0 version) soon, so that we have extensive testing during the next few months before it hits a release.

ncharles:
The file is created in distributePolicy/1.0/propagatePromises.cf , when tools are copied on the rudder root server

ncharles:
replaced by https://github.com/Normation/rudder-techniques/pull/735

PR #625, opened 2014-12-31
Fixes #6035: Do not install curl in initial promises

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6035

To be merged in 2.10


jooooooon:
@Kegeruneku It may well be already installed, but the problem could be that any promise that runs yum causes this deadlock, because it's trying to run yum, from cf-agent, which is already running within yum... Thus a locking issue.

It may just be that curl is the first package we check, no?

PR #627, opened 2015-01-07
Fixes #6089: enforce file content post modification hook if section is modified but global state of the file is not changed

peckpeck:
http://www.rudder-project.org/redmine/issues/6089


peckpeck:
Ticket rejected

PR #628, opened 2015-01-07
Fixes #6094: Corect #6089 on branch 3.0, technique 7.0

peckpeck:
http://www.rudder-project.org/redmine/issues/6094


peckpeck:
Ticket rejected

PR #629, opened 2015-01-08
Fixes #5991: reorder the check of logs system to avoid missed config upd...

ncharles:
...ate on the policy server


peckpeck:
OK, merging

PR #634, opened 2015-02-08
Fixes #6237: Wrong reporting in Technique Enforce a file content if the ...

ncharles:
...file is not present


Kegeruneku:
Good.

PR #635, opened 2015-02-09
Fixes #6238: Wrong reporting in Tehcnique Enforce a file content if the ...

ncharles:
...file is not present - Branch 3.0


Kegeruneku:
Good, linked to #634 (same logic)

PR #636, opened 2015-02-17
Fixes #6273: Fatal error in  rpmPackageInstallation/7.0 directive

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6273


peckpeck:
Good !

PR #637, opened 2015-02-18
Fixes #6279: it's not possible to define variables with " in them in gen..

ncharles:
...eric variable definition


peckpeck:
OK good

PR #638, opened 2015-03-03
Fixes #6311: Missing reports when editing twice the same file with enforce a file content v7.0

peckpeck:
http://www.rudder-project.org/redmine/issues/6311


peckpeck:
PR updated

ncharles:
thank you

PR #639, opened 2015-03-05
Fixes #6272: MOTD directive has 2 different technique versions

peckpeck:
http://www.rudder-project.org/redmine/issues/6272


peckpeck:
@ncharles ping

ncharles:
Oh,i missed it completly. Sorry.

PR #641, opened 2015-03-09
Fixes #6374: OpenSSH server technique tries to configure MaxSession on A...

ncharles:
...IX 5.3, but this parameter is invalid on this OS


Kegeruneku:
Good, according to http://en.wikipedia.org/wiki/IBM_AIX

PR #643, opened 2015-03-12
Fixes #6385: Remove /var/log/rudder/compliance/non-compliant-reports.log rotation from logrotate

fanf:
Removing in 3 places. 
Be carefull, the up-merge will imply changing in other place (node-roles)


fanf:
This ticket must be merge at the same moment than  https://github.com/Normation/rudder/pull/837

ncharles:
ok for me

PR #645, opened 2015-03-18
Fixes #6424: Correct and harmonize check-rsyslog-version and check-zyppe...

Kegeruneku:
...r-version

Ticket: http://www.rudder-project.org/redmine/issues/6424

To be merged in 2.10


tools/check-rsyslog-version
```diff
@@ -0,0 +1,26 @@
...
+RPMVERCMP="/var/rudder/cfengine-community/bin/rpmvercmp"
+
+# Minimum supported Rsyslog version
+MINIMAL_RSYSLOG_VERSION="5.7.1"
+
+if type rsyslogd >/dev/null 2>&1; then
+
+  CURRENT_RSYSLOG_VERSION=$(rsyslogd -v | head -n1 | sed "s/^rsyslogd \\([^, ]*\\).*$/\\1/")
+
+  if ${RPMVERCMP} ${MINIMAL_RSYSLOG_VERSION} lt ${CURRENT_RSYSLOG_VERSION}; then
```
ncharles: could you check the semantic here of lt ?

tools/check-zypper-version
```diff
@@ -0,0 +1,29 @@
...
+RPMVERCMP="/var/rudder/cfengine-community/bin/rpmvercmp"
+
+# Minimum supported Zypper version
+MINIMAL_ZYPPER_VERSION="1.0.8"
+
+if type zypper >/dev/null 2>&1; then
+
+  CURRENT_ZYPPER_VERSION=$(/usr/bin/zypper --version | cut -d' ' -f2)
+
+  if ${RPMVERCMP} ${MINIMAL_ZYPPER_VERSION} lt ${CURRENT_ZYPPER_VERSION}; then
```
jooooooon: What happens if they are equal? Don't you want "le" rather than "lt"?

tools/check-zypper-version
```diff
@@ -0,0 +1,29 @@
...
+    echo "+zypper_version_ok"
+  else
+    echo "+zypper_version_not_ok"
```
ncharles: you changed the name of the class ? it is however used in several techniques

ncharles: we should not change on a stable version the classes define, people may have use it in their own technique

Kegeruneku: Well, let's output both classes then :)

tools/check-zypper-version
Kegeruneku: There is no le, I'll have to add another condition for equality... :)

Kegeruneku: Or as @ncharles suggested, just reverse the condition :)

jooooooon:
Aside from the inline comments, this is an AWESOME improvement. I love seeing cleanups like this that make code simpler, and renaming files to match our convention.

Kegeruneku:
Thanks :D

Kegeruneku:
PR updated according to review !

peckpeck:
I don't see in your PR that you reverted the condition, nor that you included the old classes in the output. 
Is this the last version ?

Kegeruneku:
Yup, it is :)

Kegeruneku:
Old one was:
```
if ${RPMVERCMP} ${MINIMAL_ZYPPER_VERSION} lt ${CURRENT_ZYPPER_VERSION}; then
```

new one is:
```
if ${RPMVERCMP} ${CURRENT_ZYPPER_VERSION} lt ${MINIMAL_ZYPPER_VERSION}; then
```

Kegeruneku:
Also, I added:
```
# Also output the legacy variable name, just in case
echo "+zypper_version_unusable"
```

The other tool outputs "+rsyslog_greater_than_5_7_1" which is way too specific and IMHO, there is no need to keep the legacy ones is it is purely internal (promises.st)

peckpeck:
That's OK for me

PR #646, opened 2015-03-31
Fixes #6439: Stuck waiting for user input during log compression

peckpeck:
http://www.rudder-project.org/redmine/issues/6439


ncharles:
Clever !

PR #648, opened 2015-04-05
Fixes #6465: Update rudder_common_report and rudder_common_reports_generic to be compatible wiht multi instance technique

ncharles:


ncharles:
oops, wrong target

PR #651, opened 2015-04-13
Fixes #6501: reports/all.log logs everything, not just rudder logs

peckpeck:
http://www.rudder-project.org/redmine/issues/6501


ncharles:
This wont work: messages from the remote agent are not in local6
this change won't log them ..
You'll want to check by program name, like in what is done at the end of the file
:programname, contains, "rudder"  /var/log/rudder/reports/all.log
:msg, ereregex, "from .*: rudder:"  /var/log/rudder/reports/all.log

peckpeck:
Yes they are, why wouldn't they be ?
All rudder messages are in local6, whether from local or remote agent.

jooooooon:
This looks like a nice fix to have - @peckpeck could you rework it for master branch (I don't think we should make a behavioural change like this in a released version) and get @amousset to review it please?

peckpeck:
PR replaced by https://github.com/Normation/rudder-techniques/pull/1185

PR #654, opened 2015-04-19
Fixes #6507: rsyslog configuration should use IP based authorization

peckpeck:
http://www.rudder-project.org/redmine/issues/6507

### DIFF
--- a/techniques/system/distributePolicy/1.0/rudder-rsyslog-root.st
+++ b/techniques/system/distributePolicy/1.0/rudder-rsyslog-root.st
@@ -70,3 +70,25 @@ $template RudderReportsFormat,"insert into RudderSysEvents (executionDate, nodeI
 
 ## Syslog messages from "rudder" forwarded from AIX
 :msg, ereregex, "from .*: rudder" ~
+
+
+# Access Control list, localhost + all known agents 
+$AllowedSender UDP, 127.0.0.0/8
+$AllowedSender TCP, 127.0.0.0/8
+
+&if(SKIPIDENTIFY)&
+
+&AUTHORIZED_NETWORKS:{net|
+$AllowedSender UDP, &net&
+$AllowedSender TCP, &net&
+}&
+
+&else&
+
+&MANAGED_NODES_IP : {ip |
+$AllowedSender UDP, &ip&
+$AllowedSender TCP, &ip&
+} &
+
+&endif&
+
--- a/initial-promises/node-server/common/1.0/rudder_lib.cf
+++ b/initial-promises/node-server/common/1.0/rudder_lib.cf
@@ -204,24 +204,52 @@ body action longjob
 
 body edit_defaults noempty_backup
 {
-        empty_file_before_editing => "false";
-        edit_backup => "timestamp";  # we want to keep a track of everything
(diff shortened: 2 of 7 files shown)


techniques/system/distributePolicy/1.0/rudder-rsyslog-root.st
```diff
@@ -70,3 +70,17 @@ $template RudderReportsFormat,"insert into RudderSysEvents (executionDate, nodeI
...
+
+&if(SKIPIDENTIFY)&
+
+$AllowedSender UDP, &ALLOWEDNETWORK&
+$AllowedSender TCP, &ALLOWEDNETWORK&
+
+&else&
+
+&MANAGED_NODES_IP : {ip |
+$AllowedSender UDP, &ip&
```
ncharles: the line can get pretty big there (8 Mo if I understand correctly), or is it one line per ip ?
In the first case, how does the agent handles it (line > 4Ko are usually truncated). In the second case, it is perfect

techniques/system/distributePolicy/1.0/rudder-rsyslog-root.st
```diff
@@ -70,3 +70,17 @@ $template RudderReportsFormat,"insert into RudderSysEvents (executionDate, nodeI
 
 ## Syslog messages from "rudder" forwarded from AIX
 :msg, ereregex, "from .*: rudder:" ~
+
+&if(SKIPIDENTIFY)&
+
+$AllowedSender UDP, &ALLOWEDNETWORK&
+$AllowedSender TCP, &ALLOWEDNETWORK&
```
ncharles: what happens if there are several networks ??

peckpeck: Good point, I don't know

techniques/system/distributePolicy/1.0/rudder-rsyslog-root.st
peckpeck: It's one line per IP, hence the syntax "&MANAGED_NODES_IP : {ip |" which breaks an array into components.

techniques/system/distributePolicy/1.0/rudder-rsyslog-root.st
ncharles: you ought to test (as a matter of fact, i know that it will output 
192.168.0.42/12192.168.42.0/31 (all texts stuck))

fanf:
I think the metadata.xml of distribute-policy for the non-initial promises must be changed, to, to add the new system variable in use.

ncharles:
Indeed, the metadata should be updated

techniques/system/distributePolicy/1.0/rudder-rsyslog-root.st
peckpeck: &ALLOWEDNETWORK& is empty and in fact never used.

The correct variable to use is &AUTHORIZED_NETWORKS& and is a proper array

peckpeck:
PR updated

peckpeck:
PR updated

ncharles:
looks good to me

PR #658, opened 2015-04-27
Fixes #6515: Add OpenSSL for Windows in tools

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6515

To be merged in 3.1


initial-promises/node-server/common/1.0/site.cf
```diff
@@ -83,6 +83,9 @@ bundle common g
                                               "curl/libcurl.dll",
                                               "curl/libssl32.dll",
                                               "curl/libeay32.dll",
+                                              "openssl/openssl.exe",
+                                              "openssl/libeay32.dll",
```
peckpeck: Do we really need to duplicate the libraries ? They are already present for curl.

Kegeruneku: I could try to identify against which OpenSSL version is what's in the curl directory and bundle this version in a directory containing both curl and openssl, or I could just replace curl's original binaries with the most recent ones if it breaks nothing.

Kegeruneku: What do you think is the best approach ?

peckpeck: Having an up to date openssl library for both should be a good thing so i think you should put the most recent library here.
And try not to duplicate binaries since they already are pretty heavy

Kegeruneku:
Updated according to review !

Kegeruneku:
I updated cURL to a newer version

peckpeck:
OK Good

PR #660, opened 2015-04-30
Fixes #6527: Allow MaxSessions configuration on sshd on RHEL 6+ and SLES...

Kegeruneku:
... 11+

Ticket: http://www.rudder-project.org/redmine/issues/6527

To be merged in 2.11


Kegeruneku:
Note: This Technique would probably need a complete review of allowed options, there are very broad limitations on what can be applied on what OSes that might need to be softened since a lot of new OS versions were released in the meantime :)

Kegeruneku:
Please open a separate ticket for this

peckpeck:
Ok, as long as we can't detect valid options directly from openssh, this is the only solution.
Merging

PR #661, opened 2015-05-04
Fixes #6510: Inventory technique should create a signature and send it

peckpeck:
http://www.rudder-project.org/redmine/issues/6510


VinceMacBuche:
To @amousset and me this pull request is valid ! merging!

PR #668, opened 2015-05-13
Fixes #6585: Wrong initial promise file modified breaking tests in tehcniques

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6585


peckpeck:
OK Good

PR #670, opened 2015-05-15
Fixes #6588: Zypper is called without --non-interactive when filling available updates cache

amousset:
http://www.rudder-project.org/redmine/issues/6588


peckpeck:
OK good

PR #671, opened 2015-05-19
Fixes #6614: cf-serverd.cf is broken because of missing promise type

ncharles:


VinceMacBuche:
"access:" promise type should be moved outside the "if" and we shoudl laways define access to /var/rudder/cfengine-community/bin/cf-agent

ncharles:
yeah i'm on it
the bug is bigger than that (massive slowdon in promise generation)

VinceMacBuche:
:( :(

techniques/system/common/1.0/cf-served.st
```diff
@@ -64,31 +63,32 @@ bundle server access_rules
...
         admit   => { host2ip("&POLICYSERVER&"), escape("&POLICYSERVER&")  };
-&endif&
-    any::
+  &endif&
+
+    policy_server_set::
       # Allow server to remotely run the agent
       "/var/rudder/cfengine-community/bin/cf-agent"
-        admit   => { host2ip("${server_info.cfserved}"), escape("${server_info.cfserved}") };
+        admit   => { host2ip("${def.policy_server}"), escape("${def.policy_server}") };
```
ncharles: there is no reason to use a different variable than in l123

techniques/system/common/1.0/cf-served.st
```diff
@@ -118,7 +118,7 @@ bundle common def
       &AUTHORIZED_NETWORKS:{net|"&net&",}&
     };
 &endif&
-    !policy_server::
+    !policy_server.policy_server_set::
```
peckpeck: The  policy_server_set class is not defined in branch 2.10

techniques/system/common/1.0/cf-served.st
peckpeck: The initial promise cf-served.cf and all other files use server_info.cfserved shouldn't we correct cf-serverd.st  instead ?

ncharles: Oh, this part makes no sense in 2.10, as we have "${sys.workdir}"  maproot => { host2ip("&POLICYSERVER&"), escape("&POLICYSERVER&") },

techniques/system/common/1.0/cf-served.st
ncharles: indeed ...

ncharles:
this PR was replaced by https://github.com/Normation/rudder-techniques/pull/673

PR #673, opened 2015-05-19
Fixes #6614: cf-serverd.cf is broken because of missing promise type

peckpeck:
http://www.rudder-project.org/redmine/issues/6614


ncharles:
you forgot to add the condition on admit   => { host2ip("${server_info.cfserved}"), escape("${server_info.cfserved}") };

PR #674, opened 2015-05-21
Fixes #6624: Merge error in common technique, resulting in invalid code

ncharles:


peckpeck:
Good catch!

techniques/system/common/1.0/cf-served.st
```diff
@@ -24,14 +24,10 @@
...
 {
-<<<<<<< HEAD
-&if(MANAGED_NODES_NAME)&
-=======
->>>>>>> branches/rudder/2.10
   # Access rules are only defined on a policy server. Standard nodes should not share any files.
   access:
 
-&if(CLIENTSLIST)&
+&if(MANAGED_NODES_NAME)&
```
peckpeck: We didn't see that one

PR #676, opened 2015-05-27
Fixes #6676: rudder-agent is stuck, waiting for user input during log compression (on 2.10)

amousset:
http://www.rudder-project.org/redmine/issues/6676


ncharles:
:+1:

PR #678, opened 2015-06-02
Fixes #6687: bundle sendInventoryToCmdb tries to send .sign files to the endpoint

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6687


peckpeck:
OK Good

PR #681, opened 2015-06-09
Fixes #6727: Almost no reports on rudder webapp, since rsyslong configuration accet only the last node

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/6727


peckpeck:
OK Good

PR #682, opened 2015-06-10
Fixes #6721: When a node is promoted to relay, the old rsyslog conf is never replaced

ncharles:


techniques/system/distributePolicy/1.0/rsyslogConf.st
```diff
@@ -87,6 +87,13 @@ bundle agent install_rsyslogd {
         comment => "Add the rsyslog.conf.d include if not already present",
         classes => cf2_if_else("rsyslog_inc_ok" , "rsyslog_inc_failed");
 
+
+      # We need to delete, if it exists, the historical rudder-agent.conf file from before the promotio of the node to relay
```
Kegeruneku: s/promotio/promotion/ :)

ncharles:
PR updated

PR #684, opened 2015-06-11
Fixes #6566: promises.cf's \"bundle agent check_cron_daemon\" is not systemd aware

amousset:
http://www.rudder-project.org/redmine/issues/6566

### DIFF
--- a/initial-promises/node-server/promises.cf
+++ b/initial-promises/node-server/promises.cf
@@ -365,30 +365,44 @@ bundle agent check_uuid
 }
 
 #######################################################
+# Match all variants of cron proccesses
+
+body process_select cron_bin
+{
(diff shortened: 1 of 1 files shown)


amousset:
Do not merge for now, this will be improved.

amousset:
PR updated.

initial-promises/node-server/promises.cf
```diff
@@ -365,30 +365,44 @@ bundle agent check_uuid
 }
 
 #######################################################
+# Match all variants of cron proccesses
+
+body process_select cron_bin
+{
+  command => "(/usr/sbin/)?crond?( -n| -f)?";
```
amousset: This allows matching the complete COMMAND, and not only the end, allowing not to match anacron for example.

Kegeruneku:
<stamp> !

PR #689, opened 2015-06-11
Fixes #5622: Let the user choose between udp and tcp for syslog messages

peckpeck:
https://www.rudder-project.org/redmine/issues/5622


techniques/system/common/1.0/promises.st
```diff
@@ -482,7 +493,7 @@ bundle agent check_log_system
 
     !windows.rsyslogd.!policy_server::
       "/etc/rsyslog.d/rudder-agent.conf"
-        edit_line => append_if_no_lines("#Rudder log system${const.n}if $syslogfacility-text == 'local6' and $programname startswith 'rudder' then @@${server_info.cfserved}:&SYSLOGPORT&${const.n}if $syslogfacility-text == 'local6' and $programname startswith 'rudder' then ~"),
+        edit_line => append_if_no_lines("#Rudder log system${const.n}if $syslogfacility-text == 'local6' and $programname startswith 'rudder' then ${rsyslog_rule_prefix}${server_info.cfserved}:&SYSLOGPORT&${const.n}if $syslogfacility-text == 'local6' and $programname startswith 'rudder' then ~"),
```
ncharles: this is not ok: we need to remove the udp line and add the tcp line if in tcp (and other way round)
maybe we'd need a setconfigvalue or something like that ?

techniques/system/common/1.0/promises.st
```diff
@@ -448,10 +448,17 @@ bundle agent check_log_system
...
+      "syslogng_rule_prefix"  string => "tcp";
+    use_udp::
+      "rsyslog_rule_prefix"   string => "@";
+      "syslogng_rule_prefix"  string => "udp";
+
     any::
 
       "syslog_conf_comment"   string => "# Autogenerated by rudder, do not edit${const.n}";
-      "syslog_ng_conf_prefix" string => "filter f_local_rudder{facility(local6) and program(\"rudder.*\");};destination rudder_loghost {tcp(\"";
+      "syslog_ng_conf_prefix" string => "filter f_local_rudder{facility(local6) and program(\"rudder.*\");};destination rudder_loghost {${syslogng_rule_prefix}(\"";
```
ncharles: also here: if we switch, we must have only one transport defined

techniques/system/common/1.0/promises.st
ncharles: oh, empty_backup, so the file contain only this line :)

techniques/system/common/1.0/promises.st
ncharles: it is already correctly handled

ncharles:
This is correct, you can merge when the other part (rudder code) is fixed

techniques/system/common/1.0/nxlog_activation.st
```diff
@@ -53,6 +53,15 @@ bundle common nxlog_common
...
 
+    use_udp::
+      "transport_module"  string => "om_udp";
+    use_tcp::
+      "transport_module"  string => "om_tcp";
+
+  classes:
+    any::
+      "use_udp" expression => strcmp("&RUDDER_SYSLOG_PROTOCOL&", "UDP")
+      "use_tcp" expression => strcmp("&RUDDER_SYSLOG_PROTOCOL&", "TCP")
```
ncharles: i missed the missing ; :(

PR #695, opened 2015-06-12
Fixes #6749: Generate promises are invalid; as there is a typo in the…

ncharles:
… definition of TCP and UDP


peckpeck:
OK Good

PR #697, opened 2015-06-17
Fixes #6766: cronConfiguration is broken because of duplicate definition of cron_bin

amousset:
https://www.rudder-project.org/redmine/issues/6766


amousset:
PR updated.

PR #698, opened 2015-06-18
Fixes #6611: Openssh technique is missing reports

amousset:
https://www.rudder-project.org/redmine/issues/6611


peckpeck:
OK Good

PR #701, opened 2015-06-23
Fixes #6784: Editing non-existent /etc/rsyslog.d/pgsql.conf in initial promises

amousset:
https://www.rudder-project.org/redmine/issues/6784


Kegeruneku:
Accepted !

PR #703, opened 2015-06-25
Fixes #6831: Use HTTP for inventory handling on AIX to prevent minicurl errors

Kegeruneku:
https://www.rudder-project.org/redmine/issues/6831


peckpeck:
OK Good

PR #704, opened 2015-06-26
Fixes #6843: minicurl on AIX tries to use a function that is not defined due to OpenSSL absence

Kegeruneku:
https://www.rudder-project.org/redmine/issues/6843


peckpeck:
OK Good

PR #705, opened 2015-06-26
Fixes #6842: Edit only one syslog configuration at a time

Kegeruneku:
https://www.rudder-project.org/redmine/issues/6842

### DIFF
--- a/techniques/system/common/1.0/promises.st
+++ b/techniques/system/common/1.0/promises.st
@@ -454,9 +454,14 @@ bundle agent check_log_system
   classes:
 
     !android::
+
       "rsyslogd" expression => fileexists("/etc/rsyslog.conf");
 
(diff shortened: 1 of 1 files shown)


peckpeck:
What if someone has an old configuration file that is not used because they have migrated from a daemon to another one ?

Kegeruneku:
That's precisely the case I'm trying to solve: we want only one active configuration. rsyslog is becoming the de-facto standard, that's why I give it the top priority. Do you think I should use another indicator instead ?

peckpeck:
I think there is no problem at all. Testing the existence of a file is almost free whereas you don't know the reason a user may have to change his syslog daemon. He can go from rsyslog to syslog-ng if he needs a new feature.
So I don't see any gain in doing this.

PR #706, opened 2015-06-26
Fixes #6845: Nodes using syslogd coninue to report to their old server when changing their policy server

Kegeruneku:
https://www.rudder-project.org/redmine/issues/6845

### DIFF
--- a/techniques/system/common/1.0/promises.st
+++ b/techniques/system/common/1.0/promises.st
@@ -790,17 +790,16 @@ bundle edit_line enforce_content(str)
 bundle edit_line fix_syslogd(syslogd)
 {
   delete_lines:
-    any::
-      "^(local6)\s+(?!${syslogd}).*"
-        comment => "Delete missconfigured rudder syslogd destination";
-      "^local6\.\*\s+${syslogd}"
-        comment => "Delete old rudder syslogd format";
+
+      "^\s*local6.*"
+        comment => "Delete all local6 facility related lines";
 
   insert_lines:
-    any::
+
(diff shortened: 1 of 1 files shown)


techniques/system/common/1.0/promises.st
```diff
@@ -790,17 +790,16 @@ bundle edit_line enforce_content(str)
 bundle edit_line fix_syslogd(syslogd)
 {
   delete_lines:
-    any::
-      "^(local6)\s+(?!${syslogd}).*"
-        comment => "Delete missconfigured rudder syslogd destination";
-      "^local6\.\*\s+${syslogd}"
-        comment => "Delete old rudder syslogd format";
+
+      "^\s*local6.*"
```
peckpeck: That means we take over all local6 configuration of the user whatever it is.
Isn't it possible that they already use it for something else than rudder ?

Kegeruneku: Possibly. But it means that in this case they would also be jamming the Rudder server with useless (for us) logs if their utility outputs logs in local6.notice.

There are plenty of other facilities to use, I prefer to block logs and make things clear (they will notice the local6.* gets removed) that letting things happen an provoke hard to debug situations.

peckpeck: 8 is not plenty, and some product (like Rudder) don't let the user choose the facility
Moreover the user won't notice until there is a problem. So I'm asking if there is not another solution. And if not we should make it more visible to the user.

peckpeck:
OK Good

PR #710, opened 2015-07-01
Fixes #6904: Ubuntu trusty does not have an /etc/motd by default

amousset:
https://www.rudder-project.org/redmine/issues/6904


VinceMacBuche:
Great ! merging

PR #712, opened 2015-07-03
Fixes #6833: root_component_check tries to start at boot uneeded services on 3.1

Kegeruneku:
https://www.rudder-project.org/redmine/issues/6833


Kegeruneku:
I decided to:
* Remove bootstart checking from the initial promises (it is not their job... and they get replaced as soon as the webapp deploys for the first time)
* Rename "initscript" to "service"

Kegeruneku:
Also, I don't think reporting is necessary for the part where we set the "rudder" service at boot.

Kegeruneku:
PR updated

PR #713, opened 2015-07-03
Fixes #6929: Create lists of files from ncf to include safely (avoid …

Kegeruneku:
…4K limit and windows regexp weirdness

Ticket: http://www.rudder-project.org/redmine/issues/6929

To be merged in 2.10


Kegeruneku:
Tests pending !

ncharles:
this should be targetted to branch 2.11, we don't do this kind of modification in old stable branch, and windows is not really supported in 2.10

Kegeruneku:
PR replaced by https://github.com/Normation/rudder-techniques/pull/714

PR #719, opened 2015-07-09
Fixes #6942: Reporting on openssh server configuration technique is n…

ncharles:
…ot correct for port edition


peckpeck:
OK Good

PR #720, opened 2015-07-10
Fixes #6978: Enforce file content Technique doesn't properly migrate …

ncharles:
…lines from outside the zone to the zone (branch 3.0)


peckpeck:
OK Good

PR #727, opened 2015-07-27
Fixes #7038: When promises could have been updated, we still have the…

ncharles:
… error messages saying we couldn't update promises

### DIFF
--- a/initial-promises/node-server/common/1.0/update.cf
+++ b/initial-promises/node-server/common/1.0/update.cf
@@ -318,11 +318,14 @@ body action u_background
 ############################################
 body classes success(if, else, kept)
 {
-        promise_kept => { "${kept}" };
-        promise_repaired => {"${if}"};
-        repair_failed => { "${else}" };
-        repair_denied => { "${else}" };
-        repair_timeout => { "${else}" };
+        promise_kept     => { "${kept}" };
+        promise_repaired => { "${if}"   };
+        repair_failed    => { "${else}" };
+        repair_denied    => { "${else}" };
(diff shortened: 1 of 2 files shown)


initial-promises/node-server/common/1.0/update.cf
```diff
@@ -318,11 +318,14 @@ body action u_background
...
-        repair_denied => { "${else}" };
-        repair_timeout => { "${else}" };
+        promise_kept     => { "${kept}" };
+        promise_repaired => { "${if}"   };
+        repair_failed    => { "${else}" };
+        repair_denied    => { "${else}" };
+        repair_timeout   => { "${else}" };
+
+        cancel_kept      => { "${else}" };
+        cancel_repaired  => { "${else}" };
```
peckpeck: Shouldn't there be "${if}" in cancel_kept and "${kept}" too in cancel_repaired ?

ncharles: I'm not sure it would be relevant. If we cancel the repaired class as well, we'll most likely hide the update that occured, and user may scratch their head on why promises are updated, but report kept

peckpeck: So we'll report both a kept and repaired status ?

ncharles: no, the code is clever, and report the worse state (repaired in case of success and repaired, error in case of repaired and error, etc)

peckpeck: OK

peckpeck:
Merging

PR #728, opened 2015-07-30
Fixes #7057: On Windows, Rudder specific inventories is not be added …

ncharles:
…to inventory with most version of CFEngine Enterprise


Kegeruneku:
OK §

PR #729, opened 2015-07-31
Fixes #7063: Rudder specific information are not added anymore on the…

ncharles:
… inventory when using FusionInventory < 2.3


peckpeck:
OK Good

PR #731, opened 2015-08-13
Fixes #7098: When syslog-ng is replaced by rsyslogd on SuSE, service is not restarted

amousset:
https://www.rudder-project.org/redmine/issues/7098


amousset:
WIP

amousset:
This code fixes the problem. It probably misses some reporting, and only the check_log_system case was tested.

Kegeruneku:
Thanks Alexis. I'm taking over :)

PR #733, opened 2015-08-17
Fixes #7109: After an upgrade to 3.1.1-1, the nodes report error on \"Could not retrieve the UUID of the policy server\"

amousset:
https://www.rudder-project.org/redmine/issues/7109


Kegeruneku:
Looks perfect

PR #735, opened 2015-08-23
Fixes #5580: check the monotony of timestamp when doing the upgrades

ncharles:


techniques/system/common/1.0/failsafe.st
```diff
@@ -79,7 +79,12 @@ bundle common g
       "rudder_tools_updated_origin" string => "${rudder_tools_origin}/rudder_tools_updated";
       "rudder_tools_updated"        string => "${rudder_tools}/rudder_tools_updated";
 
-    # The time at which the execution started
+
+      "rudder_tools_updated_filename" string => "rudder_tools_updated";
+      "rudder_tools_updated_origin" string => "${rudder_tools_origin}/${rudder_tools_updated_filename}";
+      "rudder_tools_updated"        string => "${rudder_tools}/${rudder_tools_updated_filename}";
+
+     # The time at which the execution started
```
jooooooon: Minor detail, but this indendation is wrong (5 spaces - should likely be 6)

techniques/system/common/1.0/update.st
```diff
@@ -143,13 +143,64 @@ bundle agent update_action
...
+       "rudder_tools_updated_tmp"   string => "${g.rudder_tools_updated_filename}.tmp";
+
+     current_promise_timestamp_defined::
+       "current_promises_timestamp" string => readfile("${client_inputs}/${rudder_promises_timestamp}", 2048);
+     current_tools_timestamp_defined::
+       "current_tools_timestamp"   string => readfile("${g.rudder_tools_updated}", 2048);
+
+     # File ought to be there only in second pass, but since classes are persistent, we need to 
+     # enforce the check on pass_2
+     pass2.(rudder_promises_generated_tmp_file_kept|rudder_promises_generated_tmp_file_repaired)::
```
jooooooon: Shouldn't this check also say ".!rudder_promises_generated_tmp_file_error" ?

techniques/system/common/1.0/update.st
```diff
@@ -143,13 +143,64 @@ bundle agent update_action
...
+       "target_promises_timestamp" string => readfile("${client_inputs}/${rudder_promises_timestamp_tmp}", 2048);
+
+     pass2.(rudder_tools_updated_repaired|rudder_tools_updated_kept)::
```
jooooooon: Same question here

jooooooon:
This looks good to me, apart from those 2 questions.

techniques/system/common/1.0/update.st
ncharles: This is an excellent question.
Classes are persistent, and error classes are canceled in case of success; so it's probably unlikely to have error and repaired at the same time.
However, to be on the safe side; i'll add the .!error condition

ncharles:
PR updated !

jooooooon:
Thanks! Looks good to me :+1:

PR #737, opened 2015-08-27
Fixes #7086: Promise copy is not convergent: if the copy is interupte…

ncharles:
…d, it will never recover

### DIFF
--- a/techniques/system/common/1.0/update.st
+++ b/techniques/system/common/1.0/update.st
@@ -200,6 +200,12 @@ bundle agent update_action
         action       => immediate,
         classes      => success("config", "no_update", "config_ok");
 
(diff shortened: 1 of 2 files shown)


techniques/system/common/1.0/update.st
```diff
@@ -328,6 +340,9 @@ body classes success(if, else, kept)
         repair_denied    => { "${else}" };
         repair_timeout   => { "${else}" };
 
+        cancel_kept      => { "${else}" };
+        cancel_repaired  => { "${else}" };
+
```
ncharles: this is already in branch 2.11

Kegeruneku:
Accepted

PR #738, opened 2015-08-27
Fixes #7138: System technique distributePolicy is not system and has two system var section

fanf:
https://www.rudder-project.org/redmine/issues/7138


ncharles:
Looks good to me ! Thank you Francois

PR #741, opened 2015-09-02
Fixes #7150: Environnement variables cannot be used on windows

ncharles:


techniques/system/common/1.0/environment-variables.st
```diff
@@ -36,14 +36,14 @@ for /F  \"tokens=1,2* delims==\" %%G IN ('SET') DO ECHO =node.env[%%G]=%%H";
 
   files:
     !windows::
-      "${sys.workdir}/module/env"
+      "${sys.workdir}/modules/env"
```
ncharles: ${sys.workdir}/modules is the real path preexisting ...

Kegeruneku:
Accepted

PR #743, opened 2015-09-21
Fixes #7212: send-clean.sh always wait for expiration time for signed inventories

fanf:
https://www.rudder-project.org/redmine/issues/7212


peckpeck:
Ok Good

PR #744, opened 2015-09-23
Fixes #7224: Add missing system vars in inventory system technique

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7224


peckpeck:
OK Good

PR #747, opened 2015-09-29
Fixes #7235: Inventory directory are moved via send-clean.sh, making unable to receive more inventories

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7235


peckpeck:
OK Good

PR #751, opened 2015-10-08
Fixes #7255:  Ssh key distribution cannot distibute several keys on o…

VinceMacBuche:
…ne user (branch 2.11, versions 2.0 and 3.0)


peckpeck:
OK Good

PR #752, opened 2015-10-09
Fixes #7211: System technique's curl should follow redirects

peckpeck:
https://www.rudder-project.org/redmine/issues/7211


Kegeruneku:
Accepted

PR #753, opened 2015-10-09
Fixes #7264: technique's curl should follow redirects

peckpeck:
https://www.rudder-project.org/redmine/issues/7264


Kegeruneku:
OK !

PR #754, opened 2015-10-12
Fixing #7268 for apache24 network list

matya:
Make sure the allow list only consists of IP-s, otherwise it will break apache 2.4's Require IP statements.


peckpeck:
Good, thanks!

PR #755, opened 2015-10-13
Fixes #7274: Remove periodic cf-execd and cf-serverd restart at 05:00

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/7274

To be merged in 2.10


Kegeruneku:
Note: I think this is the wrong approach, we should just remove this bundle altogether (cf http://www.rudder-project.org/redmine/issues/6118 )

Kegeruneku:
Should I remove it instead ?

Kegeruneku:
I chose to merge the cron setup altogether in one single file (KISS)

peckpeck:
OK Good

PR #757, opened 2015-10-15
Fixes #7206: Touch file on inventory sent

matya:
This fix creates and keeps a file updated whenever an inventory
is successfully sent. This can be used to set up a monitor on
the node, that will check the age of the file not older then
a certain threshold.

### DIFF
--- a/initial-promises/node-server/inventory/1.0/fusionAgent.cf
+++ b/initial-promises/node-server/inventory/1.0/fusionAgent.cf
@@ -517,6 +517,11 @@ bundle agent sendInventory
         classes      => if_else("inventory_file_deleted", "cant_delete_inventory_file"),
         comment      => "Cleaning up inventory files already sent to the server";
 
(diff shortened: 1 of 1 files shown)


initial-promises/node-server/inventory/1.0/fusionAgent.cf
```diff
@@ -517,6 +517,11 @@ bundle agent sendInventory
         classes      => if_else("inventory_file_deleted", "cant_delete_inventory_file"),
         comment      => "Cleaning up inventory files already sent to the server";
 
+      "${rudder_var_tmp}/inventory_sent"
```
peckpeck: This should be g.rudder_var_tmp

peckpeck:
Aside of that everything is good.

matya:
Jeah, just noticed and pushed it. should be good now.

peckpeck:
Good, thanks!

PR #758, opened 2015-10-17
Fixes #7289: Redirect zypper's version to stdout

matya:
Zypper displays the version on stderr on very low versions.
Redirect stderr to stdout to ensure value is catched.


peckpeck:
Thank you!

PR #763, opened 2015-11-09
Implements #6756 Technique to manage Repo GPG Keys

matya:
The Technique allows multiple GPG keys to be
imported or removed from the managed systems.
It knows RPM and APT Based Repository Manager.


matya:
Reworking

PR #764, opened 2015-11-09
Implements #6756 Technique to manage Repo GPG Keys

matya:
The Technique allows multiple GPG keys to be imported or removed from the managed systems.
It knows RPM/Zypper and APT Based Repository Managers.

### DIFF
--- a/techniques/applications/repoGpgKeyManagement/1.0/metadata.xml
+++ b/techniques/applications/repoGpgKeyManagement/1.0/metadata.xml
@@ -1,0 +1,125 @@
+<!--
+
+ Copyright (c) 2015 Janos Mattyasovszky
+
+ This program is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+ Based on the Template:
+  https://github.com/Normation/rudder-techniques/blob/2c4e7fb50f0f35d786f253ca446f855b91a9ff2e/technique-metadata-sample.xml
+
+-->
+
+<TECHNIQUE name="Repository GPG Key Management for RPM / APT Systems">
+
+  <!-- This description is displayed in detailed views in the interface.
+     It should be used to describe what this Technique does, in detail.
+     New lines will be ignored. -->
+  <DESCRIPTION>
+   Manages trusted GPG Keys for the RPM / zypper / APT Package Managers.
+   It uses `rpm --import' and `apt-key add' to load, and `rpm -e' and `apt-key remove' to remove keys.
+   You can use multiple instances of Directives based on this Technique, but keep in mind:
+   when conflicting actions are specified on the same hash by different directives, it will flip-flop!
+  </DESCRIPTION>
+
+  <!-- Can several Directives based on this Technique be applied to the same node?
+     If so, the Technique will have to be written to support multi-valued variables.
+     -->
+  <MULTIINSTANCE>true</MULTIINSTANCE>
+
+  <!-- List of .st files (also called templates or TMLS)
+     in this Technique to import and parse variables in.
+     Note: the ".st" extension will be added automatically, don't specify it here -->
+  <TMLS>
+    <TML name="repoGpgKeyManagement"/>
+  </TMLS>
+
+  <!-- List of CFEngine bundles in the above .st TMLS to call.
+     These will be added to the CFEngine bundlesequence.
+       Technically, they don't have to have the same name as the
+     TML, but if there is one bundle per TML, it is recommended. -->
+  <BUNDLES>
+    <NAME>check_repo_gpg_key</NAME>
+  </BUNDLES>
+
+  <!-- Define agent and OS compatibility for this Technique.
+     Currently, this is for information purposes only.
+     ## TODO: It has not been tested on any older OS, only on latest-greatest (SLES11, CentOS 7, Ubuntu 14)
+  <COMPATIBLE>
+    <OS version=">= 5">RHEL</OS>
+    <OS version=">= 11">SuSE Linux</OS>
+    <OS version=">= 4 (Etch)">Debian</OS>
+    <OS version=">= 5">RHEL / CentOS</OS>
+  </COMPATIBLE>
+-->
+
+  <TRACKINGVARIABLE>
+    <SAMESIZEAS>GPG_KEY_HASH</SAMESIZEAS>
+  </TRACKINGVARIABLE>
+
+  <!-- From here on, define variables to display in the web interface
+     All variables must be contained in a section.
+     Sections may be multivalued, or not -->
+ <SECTIONS>
+   <SECTION name="Repository GPG Key Management" multivalued="true" component="true" componentKey="GPG_KEY_HASH">
+     <SELECT1>
+       <NAME>GPG_KEY_ACTION</NAME>
+       <DESCRIPTION>Which operation should be done with this GPG Key</DESCRIPTION>
+       <ITEM>
+         <LABEL>Import (both hash and key content required)</LABEL>
+         <VALUE>add</VALUE>
+       </ITEM>
+       <ITEM>
+         <LABEL>Remove (only hash is required)</LABEL>
+         <VALUE>del</VALUE>
+       </ITEM>
+       <CONSTRAINT>
+         <DEFAULT>add</DEFAULT>
+       </CONSTRAINT>
+     </SELECT1>
+     <INPUT>
+       <NAME>GPG_KEY_HASH</NAME>
+       <CONSTRAINT>
+         <MAYBEEMPTY>false</MAYBEEMPTY>
+         <REGEX error="Exactly 16 hexadecimal characters required"><![CDATA[ [a-fA-F0-9]{16} ]]></REGEX>
+       </CONSTRAINT>
+       <DESCRIPTION>Long hash of the GPG Key</DESCRIPTION>
+       <LONGDESCRIPTION>
+        You get it by looking for a line like `pub   2048R/70AF9E8139DB7C82 2013-01-31' when using the command `gpg --list-keys --keyid-format=long'.
+        From that, fill in the 16 hexa-chars part like `70AF9E8139DB7C82' in the example.
+       </LONGDESCRIPTION>
+     </INPUT>
+     <INPUT>
+       <NAME>GPG_KEY_NAME</NAME>
+       <CONSTRAINT>
+         <MAYBEEMPTY>true</MAYBEEMPTY>
+         <REGEX error="No double quotation marks"><![CDATA[ [^"]* ]]></REGEX>
+       </CONSTRAINT>
+       <DESCRIPTION>Description of Key</DESCRIPTION>
+       <LONGDESCRIPTION>This is only used to identify the Key in the GUI, it has no role in adding/removing the Key on the systems</LONGDESCRIPTION>
+     </INPUT>
+     <INPUT>
+       <NAME>GPG_KEY_CONTENT</NAME>
+       <CONSTRAINT>
+         <MAYBEEMPTY>true</MAYBEEMPTY>
+         <TYPE>textarea</TYPE>
+         <REGEX error="Need the whole part of PGP PUBLIC KEY BLOCK, as exported by `gpg -a --export'">
+           <![CDATA[ \s*-----BEGIN PGP PUBLIC KEY BLOCK-----[^'"\\]+-----END PGP PUBLIC KEY BLOCK-----\s* ]]>
+         </REGEX>
+       </CONSTRAINT>
+       <DESCRIPTION>The whole section of PGP PUBLIC KEY BLOCK (Only required when Importing a key)</DESCRIPTION>
+       <LONGDESCRIPTION>Use the command `gpg -a --export' to export the public key and paste the whole output here</LONGDESCRIPTION>
+     </INPUT>
+ </SECTION>
+</SECTIONS>
+</TECHNIQUE>
--- a/techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
+++ b/techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
@@ -1,0 +1,239 @@
+#####################################################################################
+#
+#  Copyright (c) 2016 Janos Mattyasovszky
+#
+#  This program is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+bundle agent check_repo_gpg_key
+{
+
+  vars:
+
+      &GPG_KEY_ACTION:{action |"repo_gpg_action[&i&]" string => "&action&";
+}&
+
+      &GPG_KEY_HASH,GPG_KEY_NAME:{hash,name |"repo_gpg_hash[&i&]" string => "&hash&", comment => "&name&";
+}&
+
+      &TRACKINGKEY:{uuid |"repo_gpg_uuid[&i&]" string => "&uuid&";
+}&
+
+      "keyid" slist => getindices("repo_gpg_hash");
+
+&GPG_KEY_CONTENT:{content |"repo_gpg_content[&i&]" string => "&content&";
+}&
+
+      "repo_gpg_hash_lc[${keyid}]" string => execresult("${paths.echo} ${repo_gpg_hash[${keyid}]} | ${paths.tr} \"[:upper:]\" \"[:lower:]\"","useshell"),
+        comment => "We need lower case to query RPM and it's also used in the apt-based module for class definition.";
+
+      "repo_gpg_hash_uc[${keyid}]" string => execresult("${paths.echo} ${repo_gpg_hash[${keyid}]} | ${paths.tr} \"[:lower:]\" \"[:upper:]\"","useshell"),
+        comment => "To remove a key from APT we need the hash in explicit capital letters, lower/mixed case just does not work, even if it reports OK.";
+
+      "repo_gpg_homedir"  string => "${g.rudder_var_tmp}/repo_gpg",
+        comment => "The temporary home directory of GPG, where it puts its trust.db, required to perform all queries by GPG.";
+
+      "repo_gpg_file[${keyid}]" string => "${repo_gpg_homedir}/key_${keyid}_content",
+        comment => "File to store the content of the key, used for later verification and import";
+
+      "repo_gpg_binary"   string => "/usr/bin/gpg",
+        comment => "The gpg binary used to verify the keys and enumerate the long hash from the RPM-s";
+
+      "repo_gpg_options"  string => "--homedir ${repo_gpg_homedir} --quiet --with-colon --keyid-format=long",
+        comment => "Common options to provide to either gpg reading the rpm -q's output or apt-key adv, which is forwarded to a gpg invocation";
+
+    repo_gpg_uses_rpm::
+      "repo_gpg_getallkeys" string => "${paths.rpm} -q gpg-pubkey --qf '%{description}\n' | ${repo_gpg_binary} ${repo_gpg_options}",
+        comment => "We use GPG to decode the whole PGP PUBLIC KEY BLOCK stored in %{description}, because rpm only stores the first 8 chars of the hash";
+
+    repo_gpg_uses_apt::
+      "repo_gpg_getallkeys" string => "${paths.apt_key} adv --fingerprint ${repo_gpg_options}",
+        comment => "'apt-key adv' is an invocation of gpg directly, so we can use the same parameters as for rpm and parse the same output format later via pipe";
+
+  classes:
+
+      "repo_gpg_key_${keyid}_action_remove"
+        expression => strcmp("${repo_gpg_action[${keyid}]}","del"),
+        comment    => "We defined this class to set the requested action for this key: Remove it from the trusted keys";
+
+      "repo_gpg_key_${keyid}_action_add"
+        expression => strcmp("${repo_gpg_action[${keyid}]}","add"),
+        comment    => "We defined this class to set the requested action for this key: Import it as a trusted key";
+
+      "repo_gpg_has_gpg_binary"
+        expression => isexecutable("${repo_gpg_binary}"),
+        comment    => "This defines if we have an executable gpg binary, necessary for gpg validation or enumerate the gpg keys on an RPM system";
+
+      "repo_gpg_has_awk_binary"
+        and     => { isvariable("paths.awk"), isexecutable("${paths.awk}") },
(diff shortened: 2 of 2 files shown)


techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,135 @@
...
+    "${g.rudder_var_tmp}/gpg_key_${index_repo_gpg_keys}_content"
+      comment    => "Remove temporary file if the GPG key has been imported successfully",
+      delete     => tidy,
+      ifvarclass => "repo_gpg_key_${index_repo_gpg_keys}_imported";
+
+  commands:
+
+   debian|ubuntu::
+
+     "/usr/bin/apt-key fingerprint | sed -rn 's/^pub\s+.{5}\/([A-F0-9]{8}) .+/\+repo_gpg_key_\1_present/gp;' | tr '[:upper:]' '[:lower:]'"
```
ncharles: Whaow ! This is impressive ! :+1:

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,135 @@
...
+    "index_repo_gpg_keys" slist => getindices("repo_gpg_hash");
+
+&GPG_KEY_CONTENT:{content |"repo_gpg_content[&i&]" string => "&content&";
+}&
+
+    ## Unfortunately string_downcase() is only available in 3.6.0, so we need to hack it with some shell
+    "repo_gpg_hash_lc[${index_repo_gpg_keys}]" string => execresult("/bin/echo ${repo_gpg_hash[${index_repo_gpg_keys}]} | tr \"[:upper:]\" \"[:lower:]\"","useshell");
+
+   debian|ubuntu::
+    "repo_gpg_hash_uc[${index_repo_gpg_keys}]" string => execresult("/bin/echo ${repo_gpg_hash[${index_repo_gpg_keys}]} | tr \"[:lower:]\" \"[:upper:]\"","useshell");
```
ncharles: either this one or the previous one is redundant

ncharles: oops, i mistread the lc/uc

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,135 @@
...
+      expression => strcmp("${repo_gpg_action[${index_repo_gpg_keys}]}","del");
+
+    "repo_gpg_key_${index_repo_gpg_keys}_action_add"
+      expression => strcmp("${repo_gpg_action[${index_repo_gpg_keys}]}","add");
+
+  files:
+    "${g.rudder_var_tmp}/gpg_key_${index_repo_gpg_keys}_content"
+      comment       => "Create a temporary file for repo to import the GPG key. Unfortunately rpm does not accept /dev/stdin or - for importing",
+      create        => "true",
+      edit_line     => set_gpg_file_content( "${repo_gpg_content[${index_repo_gpg_keys}]}" ),
```
ncharles: you could simply use insert_lines from the stdblib

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,135 @@
...
+      edit_defaults => empty_nobackup,
```
ncharles: you can use "ncf_empty_select" for that, even if the name is not really friendly.
or insert this method in the rudder_stdlib (rudder_empty_nobackup)

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,135 @@
...
+
+    "@@repoGpgKeyManagement@@result_error@@${repo_gpg_uuid[${index_repo_gpg_keys}]}@@GPG Key Management@@${repo_gpg_hash[${index_repo_gpg_keys}]}@@${g.execRun}##${g.uuid}@#The GPG Key could not be removed"
+      ifvarclass => "repo_gpg_key_${index_repo_gpg_keys}_remove_error";
+
+    "@@repoGpgKeyManagement@@result_error@@${repo_gpg_uuid[${index_repo_gpg_keys}]}@@GPG Key Management@@${repo_gpg_hash[${index_repo_gpg_keys}]}@@${g.execRun}##${g.uuid}@#The temporary file could not be created"
+      ifvarclass => "gpg_key_file_${index_repo_gpg_keys}_create_failed";
+
+}
+
+bundle edit_line set_gpg_file_content(c)
```
ncharles: you don't need this bundle edit_line, stdlib provides what is necessary

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,135 @@
...
+{
+  insert_lines:
+    "${c}";
+}
+
+body edit_defaults empty_nobackup
```
ncharles: i'm amazed that we don't already have this one on the stdlib. See my comment up there

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,135 @@
...
+
+    "/bin/rpm --quiet --import ${g.rudder_var_tmp}/gpg_key_${index_repo_gpg_keys}_content"
+      comment    => "This imports the GPG key if the necessary tmp file has been created and it needs to be imported and is not present",
+      ifvarclass => "repo_gpg_key_${index_repo_gpg_keys}_action_add.!repo_gpg_key_${repo_gpg_hash_lc[${index_repo_gpg_keys}]}_present.gpg_key_file_${index_repo_gpg_keys}_created",
+      classes    => if_else("repo_gpg_key_${index_repo_gpg_keys}_imported", "repo_gpg_key_${index_repo_gpg_keys}_import_error");
+
+  reports:
+    
+    "@@repoGpgKeyManagement@@result_repaired@@${repo_gpg_uuid[${index_repo_gpg_keys}]}@@GPG Key Management@@${repo_gpg_hash[${index_repo_gpg_keys}]}@@${g.execRun}##${g.uuid}@#The GPG Key was imported successfully" 
+      ifvarclass => "repo_gpg_key_${index_repo_gpg_keys}_imported";
```
ncharles: For the reports, you may want to use rudder_common_report (see https://github.com/Normation/rudder-techniques/blob/master/techniques/applications/aptPackageInstallation/4.0/aptPackageInstallation.st#L89 ) as it would embedeed directly support for different reporting type (change_only, full compliance..)

ncharles:
@matya This is a really nice contribution, thank you.
I made some feeback on it, mostly about consistency or using already existing methods, but the overall logic looks sound

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,135 @@
...
+      expression => strcmp("${repo_gpg_action[${index_repo_gpg_keys}]}","add");
+
+  files:
+    "${g.rudder_var_tmp}/gpg_key_${index_repo_gpg_keys}_content"
+      comment       => "Create a temporary file for repo to import the GPG key. Unfortunately rpm does not accept /dev/stdin or - for importing",
+      create        => "true",
+      edit_line     => set_gpg_file_content( "${repo_gpg_content[${index_repo_gpg_keys}]}" ),
+      edit_defaults => empty_nobackup,
+      ifvarclass    => "!repo_gpg_key_${repo_gpg_hash_lc[${index_repo_gpg_keys}]}_present.repo_gpg_key_${index_repo_gpg_keys}_action_add",
+      classes       => if_else("gpg_key_file_${index_repo_gpg_keys}_created", "gpg_key_file_${index_repo_gpg_keys}_create_failed");
```
ncharles: you also may want to consider using the classes_generic rather than if_else (if gives also the _reached)
It makes the promises easier to read (but the report a bit hard to understand though..)

matya:
@ncharles reworked technique, also fixed some bugs

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
matya: I reworked rpm to also provide classes as a module, so now missing gpg keys do not show up as  failed command executions in red, when getting colorized outputs, also the runtime changes from O(n) to O(1).

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,142 @@
...
+
+    &TRACKINGKEY:{uuid |"repo_gpg_uuid[&i&]" string => "&uuid&";
+}&
+
+    "repo_gpg_index" slist => getindices("repo_gpg_hash");
+
+&GPG_KEY_CONTENT:{content |"repo_gpg_content[&i&]" string => "&content&";
+}&
+
+    "repo_gpg_hash_lc[${repo_gpg_index}]" string => string_downcase("${repo_gpg_hash[${repo_gpg_index}]}"),
```
ncharles: oh, so this is not compatible with CFEngine 3.5 anymore :(

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,142 @@
...
+    "${g.rudder_var_tmp}/repo_gpg_key_${repo_gpg_index}_content"
+      comment    => "Remove temporary file if the GPG key import has been tried, regardless of the outcome",
+      delete     => tidy,
+      ifvarclass => "repo_gpg_key_${repo_gpg_index}_imported_reached";
+
+  commands:
+
+   _stdlib_path_exists_apt_key::
+
+    "$(paths.apt_key) fingerprint | $(paths.sed) -rn 's/^pub\s+.{5}\/([A-F0-9]{8}) .+/\+repo_gpg_key_\1_present/gp;' | $(paths.tr) '[:upper:]' '[:lower:]'"
```
ncharles: Sorry to be nitpicky, but we have the habit of using ${ } rather than $( )
could you use ${paths.apt_key}

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,142 @@
...
+
+   repo_gpg_hashes_read_not_ok::
+    "any" usebundle => repo_gpg_report("result_error", "${repo_gpg_index}", "Could not enumerate the existing keys");
+
+   !(_stdlib_path_exists_rpm|_stdlib_path_exists_apt_key)::
+    "any" usebundle => repo_gpg_report("result_na",    "${repo_gpg_index}", "Can only handle RPM and APT at the moment.");
+
+}
+
+bundle agent repo_gpg_report(status, gpg_key_id, message)
```
ncharles: this is interesting to improve readability !

ncharles:
This looks awesome ! However, could you avoid using 3.6 only features (string_downcase and string_uppercase), as it breaks compatibility with older agents

Thank you !

matya:
@ncharles done.

ncharles:
Sorry for the late review.
The code looks correct to me, we'd need to test it (and have someone with real competencies in repo public key management).
@jooooooon or @peckpeck , could you test this technique?

matya:
Any luck? :)

matya:
ping @jooooooon  @peckpeck @ncharles

jooooooon:
Sorry @matya, this appears to have slipped through the net :(

Our bad - we're just getting close to a release, so lots of bug fixes flying everywhere. We'll try and look at this this week.

matya:
Hope you all got back safely from cfgmgmtcamp and FOSDEM ;-)

matya:
@jooooooon What do you need for testing/merging? Do you need any further input or feedback?

matya:
I have identified a possible security issue, that is using the short hash, which is prone to collision-attack. I could rewrite the code the also use the long version, but that would also require the invocation of gpg in a call, as the rpmdb does only use the short hash as %{version} to identify them (well, it also has the creation unix timestamp as a hexhash in %{release}).
See:
http://security.stackexchange.com/questions/84280/short-openpgp-key-ids-are-insecure-how-to-configure-gnupg-to-use-long-key-ids-i
http://blog.andreas-haerter.com/2012/03/06/rpm-yum-gpg-key-verification-import-deletion-package-signature-check-cheat-sheet

matya:
rewrote & tested with 16-char hashes.

matya:
# Info about repo gpg key management - what is this stuff?

## Intro:
RPM or DEB Packages' catalog (metadata.xml vs Release) file is signed by a detached PGP signature, so the package manager is verify te repository at once at the beginning without having to do heavy calculations on RPM/DEB packages during installation. This is NOT having RPM/DEB **packages** being signed, it's just about the repo.
PGP Keys have a pretty long fingerprint, from which we usually use the last 32bits (8char), but 
today we can [pretty easy brute-force a colliding key id](http://security.stackexchange.com/questions/84280/short-openpgp-key-ids-are-insecure-how-to-configure-gnupg-to-use-long-key-ids-i), so we prefer to use the last 64bits,
which is implemented very good in this context, and we need the 16char-keyid for operations.

## Two worlds, to ways to store the keys:
### RPM
uses its rpmdb to store the PGP PubKeys, each pubkey is assined a virtual package name in the format of [gpg-pubkey-{shorthash}-{timeinhex}](http://blog.andreas-haerter.com/2012/03/06/rpm-yum-gpg-key-verification-import-deletion-package-signature-check-cheat-sheet
), and the PGP pubkey itself is stored in the description field of the RPM. Unfortunately RPM does not store the long hash of the key anywhere, so it has to be calculated each time we check. Good thing is this does not take too much time for a reasonable number of hashes (<20). Importing can be done by `rpm --import {pubkey_file}` - unfortunately this file has to exist, no tricks with fd-redirection or stdin(-) is accepted. Removal is simple `rpm -e` - just like deleting a package, but this requires to have the {shorthash} value. Listing all pubkeys is quite fast with `rpm -q gpg-pubkey`, it only operated on the internal rpmdb.

### APT
uses directly a private GPG keyring, which has an interface with the apt-key binary, that does all the backend gpg-command handling. Fortunately it also has the `apt-key adv` option, which takes any arguments gpg itself can handle, and passes them 1:1 to the internal gpg call. This can be used to retrieve all the long GPG hashes from its database, but unfortunately any apt-key call takes relatively long (2-5sec), so minimizing the number of calls required optimizes the runtime. Bad thing is `apt-key del` does not accept the long hash, so we need to fall back to the short and upper cased one (yeah, pretty crappy). Adding is easy, similar to rpm, by using `apt-key add {file}`.

## Under the hood:
We basically build arrays with long and short hashes, check if our system uses rpm or apt, if we are adding any files that is not added yet, we put that pubkey into a temporary file and load it, then remove the file. If we are removing keys, we fall back to the short hash of 8-chars, because that's how the apt/rpm interfaces expect the keyids. However, each check if the key is installed/absent is performed on the full-length 16char-keyid, so it is safe to assume that each positive result is a valid one. That's why we use in both cases the common `gpg --with-colon --keyid-format=long`, and then parse the machine-readable output with awk easily.

matya:
@jooooooon @ncharles @peckpeck hope this helps in the technical / functional review :)

matya:
Ping @peckpeck @jooooooon ?

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,161 @@
...
+      create        => "true",
+      edit_line     => insert_lines( "${repo_gpg_content[${repo_gpg_index}]}" ),
+      edit_defaults => ncf_empty_select("true"),
+      ifvarclass    => "repo_gpg_hashes_read_ok.!repo_gpg_key_${repo_gpg_hash_lc[${repo_gpg_index}]}_present.repo_gpg_key_${repo_gpg_index}_action_add",
+      classes       => classes_generic("repo_gpg_key_${repo_gpg_index}_file_created");
+
+    "${g.rudder_var_tmp}/repo_gpg_key_${repo_gpg_index}_content"
+      comment    => "Remove temporary file if the GPG key import has been tried, regardless of the outcome",
+      delete     => tidy,
+      ifvarclass => "repo_gpg_key_${repo_gpg_index}_imported_reached";
```
peckpeck: Why do you remove it ?
Keeping it would help debugging and avoid having to create/delete on each run which can be a bad idea on SSD.

techniques/applications/repoGpgKeyManagement/1.0/metadata.xml
```diff
@@ -0,0 +1,125 @@
...
+         <MAYBEEMPTY>true</MAYBEEMPTY> 
+         <REGEX error="No double quotation marks"><![CDATA[ [^"]* ]]></REGEX>
+       </CONSTRAINT>
+       <DESCRIPTION>Description</DESCRIPTION>
+       <LONGDESCRIPTION>This is only used to identify the Key in the GUI, it has no role in adding/removing the Key on the systems</LONGDESCRIPTION>
+     </INPUT>
+     <INPUT>
+       <NAME>GPG_KEY_CONTENT</NAME>
+       <CONSTRAINT>
+         <MAYBEEMPTY>true</MAYBEEMPTY>
```
peckpeck: Why is that optional ?
This seems to be mandatory for me as you really don't use the hash

peckpeck: anymore

peckpeck: Ok, the hash is used, but the content still seems mandatory.

peckpeck:
Aside of that this looks good to me :+1:

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
matya: I only remove it after the import step has been tried. 
1) If it succeeds, imported_reached is not set again, since Line 105 and Line 117 in repoGpgKeyManagement.st depends on that it's not imported yet and needs to be added.
2) If it fails you'd get a failure - on which I agree, could cause a repeated create+remove - and investigate hopefully and fix the problem.

techniques/applications/repoGpgKeyManagement/1.0/metadata.xml
matya: The content is only important if being imported. If you want to remove a key, you only need it's hash. It's only mandatory when being imported. Why would you need the content of the key if you only want to remove it?

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
matya: I could put it to /dev/shm, a ramdisk, the size of the files would be 2-4kBytes each for the short time of being imported. That should fit to memory on any system running deb/rpm packages and is supported to run a Rudder agent...

peckpeck: You should replace import_reached by import_ok, otherwise it will be removed even in case of error.

peckpeck: You're right about 1) it's ok for me.

techniques/applications/repoGpgKeyManagement/1.0/metadata.xml
```diff
@@ -0,0 +1,125 @@
...
+         <TYPE>textarea</TYPE>
+         <REGEX error="Need the whole part of PGP PUBLIC KEY BLOCK, as exported by `gpg -a --export'">
+           <![CDATA[ \s*-----BEGIN PGP PUBLIC KEY BLOCK-----[^'"\\]+-----END PGP PUBLIC KEY BLOCK-----\s* ]]>
+         </REGEX>
+       </CONSTRAINT>
+       <DESCRIPTION>The whole section of PGP PUBLIC KEY BLOCK</DESCRIPTION>
+       <LONGDESCRIPTION>Use the command `gpg -a --export' to export the public key</LONGDESCRIPTION>
```
peckpeck: You could add something like "mandatory when adding a key and not used when removing one"

matya: There is already a mentioning of it at the Add/Remove's <LABEL> you must choose the action at, and also at the top description, I presumed that would suffice, but I can also add mentioning it to the key's description.

matya:
Additional questions came up when I looked at the module ansible uses to manage apt key:
- do we need an additional sanity check, that the provided hash actually belongs to the key content you provided? Otherwise the Add command would run every time, since the check would always fail to find the imported key... 
- do we need to check for the presence of gpg binary? Apt always requires it because the way they use their keyring, but RPM has built in methods of handling the keys, and I could imagine you could make a stripped down sles/CentOS image than ford not have it. In that case it would be better to catch and report it then have the command promise fail the bad way...

techniques/applications/repoGpgKeyManagement/1.0/metadata.xml
peckpeck: OK, my bad

peckpeck:
It sure would be a good idea to have additional sanity checks.
It is always useful until someone has a problem, but I would understand if you don't want to do it after all the work you've already done

matya:
@peckpeck added and tested logic to verify user-entered gpg key with hash, and also added some better sanity checkings and dedicated tmp home for gpg to run (not to pollute root's ~/.gnupg dir).
Feedback / questions are welcome.

techniques/applications/repoGpgKeyManagement/1.0/repoGpgKeyManagement.st
```diff
@@ -0,0 +1,239 @@
...
+   repo_gpg_uses_apt|repo_gpg_uses_rpm::
+
+    "${repo_gpg_getallkeys} | ${paths.awk} -F ':' '$1 ~ /^pub/ { printf \"+repo_gpg_key_%s_present\n\", tolower($5); }'"
+      comment    => "We use this module instead of looking up each gpg key, because each apt-key call takes about 3-5 seconds, so this way we have O(1) instead of O(n)...",
+      ifvarclass => "repo_gpg_has_awk_binary",
+      classes    => classes_generic("repo_gpg_hashes_read"),
+      contain    => outputable,
+      module     => "true";
+
+    "${repo_gpg_binary} ${repo_gpg_options} ${repo_gpg_file[${keyid}]} | ${paths.awk} -v HASH='${repo_gpg_hash_lc[${keyid}]}' -F':' '$1 == \"pub\" { ++count; if (tolower($5) == HASH) { printf \"+repo_gpg_file_%s_validated\n\", HASH; } } END { if (count > 1) { printf \"+repo_gpg_file_%s_multikeyed\n\", HASH; } }'"
```
ncharles: Super awk-fu !

ncharles:
This looks great ! 
One last final remark (not to be nitpicking, but tests would fail otherwise)
The indentation is a bit off, please refer to http://www.rudder-project.org/foswiki/Development/TechniqueBestPractices

>  Indentation should be made by spaces. A incrementation of indentation is equal to two spaces
The promise type should be indented by two spaces (instead of being at the same indentation level than the bundle name)
The class expression should be indented by four spaces (two spaces after the promise type)
The promiser should be indented by six spaces (two spaces after the class expression or four spaces after the promise type if no class expression is defined)
Attributes of promises should be indented by eight spaces (two spaces after the promiser) and it should be only one attribute by line. 


Thank you !

matya:
@ncharles if that is all that's required, I am more than happy to comply with the need of additional spaces :-D

matya:
Btw, do you have acceptance tests for rudder techniques as well?

jooooooon:
> Btw, do you have acceptance tests for rudder techniques as well? 

There are some, mostly style / basic patterns to avoid, in https://github.com/Normation/rudder-techniques/blob/master/scripts/check-techniques.sh.

Otherwise, there are few Technique-specific tests, but all rudder-test-framework tests can and should test Technique behaviour.

matya:
Ping @ncharles  @peckpeck

ncharles:
Awesome. Merging.
Thank you for your patience

matya:
Thanks.

ncharles:
Ok, for the record, there are tests to run, in the scripts folder: check-techniques.sh
They check the normal ordering, plus that the versions are maintained

PR #765, opened 2015-11-12
Fixes #7400: Remove $() variables in cfengine code

peckpeck:
https://www.rudder-project.org/redmine/issues/7400


jooooooon:
FYI: the reason for this PR is that it breaks our style tests (no $() syntax)

PR #766, opened 2015-11-13
Bug 7379/create a new version of download file from the shared folder that would allow to set the suid and gid to multiple file

ncharles:


ncharles:
PR updated

PR #771, opened 2015-11-20
Fixes #7094 Multi-Line ENV variables parsing fixed

matya:
Not using 'command module' to enumerate env variables, as multi-line
strings are not well supported. First we get the env variable's names,
then read their values by a cfengine built-in getenv command.
 Will not work on Solaris/AIX, as their env does not support the -0
option, to output it null-terminated.

### DIFF
--- a/techniques/system/common/1.0/environment-variables.st
+++ b/techniques/system/common/1.0/environment-variables.st
@@ -30,12 +30,16 @@ bundle agent get_environment_variables
       "script" string => "#! /bin/sh
 env | sed 's/=/]=/' |sed 's/^/=node.env[/'";
 
+    !windows.linux::
+      "env_vars" slist => splitstring( execresult("(${paths.printf} '\0'; ${paths.env} -0) | ${paths.grep} -aoP '\x00([^=]+)' | ${paths.tr} -d '\000' | ${paths.tr} '\n' ',' | ${paths.sed} 's/,$//'","useshell"), "," , 2000);
+      "node.env[${env_vars}]" string => getenv( "${env_vars}", 5000);
+
     windows::
      "script" string => "@echo off
 for /F  \"tokens=1,2* delims==\" %%G IN ('SET') DO ECHO =node.env[%%G]=%%H";
 
   files:
-    !windows::
+    !windows.!linux::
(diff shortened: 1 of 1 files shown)


jooooooon:
Nice workaround, @matya ! This looks good to me, but I have one question - coming as an in-line comment.

techniques/system/common/1.0/environment-variables.st
```diff
@@ -30,12 +30,16 @@ bundle agent get_environment_variables
       "script" string => "#! /bin/sh
 env | sed 's/=/]=/' |sed 's/^/=node.env[/'";
 
+    !windows.linux::
+      "env_vars" slist => splitstring( execresult("(${paths.printf} '\0'; ${paths.env} -0) | ${paths.grep} -aoP '\x00([^=]+)' | ${paths.tr} -d '\000' | ${paths.tr} '\n' ',' | ${paths.sed} 's/,$//'","useshell"), "," , 2000);
+      "node.env[${env_vars}]" string => getenv( "${env_vars}", 5000);
```
jooooooon: Are you sure you can define variables in another context like this?

jooooooon: OK, I'm told you can, never mind :)

peckpeck:
The issue is against 3.1  but the PR is against 3.0

techniques/system/common/1.0/environment-variables.st
```diff
@@ -30,12 +30,16 @@ bundle agent get_environment_variables
       "script" string => "#! /bin/sh
 env | sed 's/=/]=/' |sed 's/^/=node.env[/'";
 
+    !windows.linux::
```
jooooooon: This could just be "linux::" :)

jooooooon:
> The issue is against 3.1 but the PR is against 3.0

The original implementation appeared in 3.0, so that's where this bug should be fixed. let's retarget the issue.

jooooooon:
Tested and approved. Thank you @matya!

techniques/system/common/1.0/environment-variables.st
matya: true, silly me :)

PR #772, opened 2015-11-22
Fixes #7430: internal security check always assume that the node is a…

ncharles:
… community one


peckpeck:
I can find references to cfengine-community in techniques/system/common/1.0/cf-served.st shouldn't it be changed too ?

ncharles:
ha, indeed...

ncharles:
PR updated

PR #774, opened 2015-11-23
Fixes #7433: Agent does not output any reports in "no reports" mode

ncharles:


jooooooon:
If we change this, then any nodes already setup with reporting will continue to output reports, even if the no-reports setting is later enabled.

I'm for the change, but we should add a "reverse" syslog config - that is, a configuration that *removes* our logging config from the local syslog installation (aka cleanup).

ncharles:
this is an interesting remark
that would indeed be relevant to be able to remove the syslog config when setting to no reports.

jooooooon:
This will be tracked in http://www.rudder-project.org/redmine/issues/7441

PR #775, opened 2015-11-23
Fixes #7439: reports-disabled is not a valid class 

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7439


techniques/system/common/1.0/rudder_stdlib_core.st
```diff
@@ -123,7 +123,7 @@ bundle agent rudder_common_report(technique_name, status, identifier, component_
 bundle agent rudder_common_report_index(technique_name, status, identifier, component_name, component_key, message, index)
 {
   classes:
-    changes_only|reports-disabled::
+    changes_only|reports_disabled::
```
ncharles: i am ashamed of that :(

PR #776, opened 2015-11-23
Fixes #7432: Don't copy files of local/50_techniques on the nodes

ncharles:


techniques/system/common/1.0/cf-served.st
```diff
@@ -60,6 +60,10 @@ bundle server access_rules
         maproot => {  @{def.acl}  },
         admit   => {  @{def.acl}  };
 
+      # Deny access to 50_techniques folder
+      "${g.rudder_ncf_origin_local}/50_techniques"
+        deny    => {  @{def.acl}  };
```
peckpeck: why not a deny all ?

ncharles: there is no all (we could put a .* ) but I'd rather stay on the safe side , since we authorize only def.acl

techniques/system/common/1.0/promises.st
```diff
@@ -124,7 +124,6 @@ bundle common va
       "ncf_local_inputs_20" slist => splitstring(execresult("${list_compatible_inputs} ${sys.cf_version} ${ncf_path}/local 20_cfe_basics", "useshell"), "\n", 10000);
       "ncf_local_inputs_30" slist => splitstring(execresult("${list_compatible_inputs} ${sys.cf_version} ${ncf_path}/local 30_generic_methods", "useshell"), "\n", 10000);
       "ncf_local_inputs_40" slist => splitstring(execresult("${list_compatible_inputs} ${sys.cf_version} ${ncf_path}/local 40_it_ops_knowledge", "useshell"), "\n", 10000);
-      "ncf_local_inputs_50" slist => splitstring(execresult("${list_compatible_inputs} ${sys.cf_version} ${ncf_path}/local 50_techniques", "useshell"), "\n", 10000);
```
jooooooon: I understand this change at the moment, because we're all working on it, but I know I won't in a few months time.

Could you replace this deleted line with "# Files from 50_techniques are automatically included by Rudder's metadata.xml mechanism" or something similar here please?

techniques/system/common/1.0/cf-served.st
peckpeck: The safe side for a deny rule is to use .*

ncharles:
pr updated

PR #778, opened 2015-11-24
Fixes #7441: Remove syslog reporting when reports are disabled

jooooooon:
https://www.rudder-project.org/redmine/issues/7441


techniques/system/common/1.0/promises.st
```diff
@@ -635,6 +638,52 @@ bundle agent check_rsyslog_version {
...
+
+  classes:
+
+      "syslogng" expression => fileexists("${check_log_system.syslogng_conffile}");
+      "syslogd"  expression => fileexists("${check_log_system.syslog_conffile}");
+
+  files:
+      "${check_log_system.rsyslog_rudder_conffile}"
+        delete         => tidy,
+        classes        => classes_generic("remove_rudder_syslog_configuration_remove_files");
```
jooooooon: This file removal could be only done if we detect rsyslog (like the 2 following blocks), but since it is most definitely our file (named rudder-agent.conf) and CFEngine considers this is a success, we might as well do it every time, just in case a node had a rsyslog server on it once - a good chance to clear up, and less code to maintain.

techniques/system/common/1.0/promises.st
```diff
@@ -635,6 +638,52 @@ bundle agent check_rsyslog_version {
...
+
+    syslogd::
+      "${check_log_system.syslog_conffile}"
+        edit_line      => delete_lines_matching("${syslog_delete_lines_patterns}"),
+        edit_defaults  => noempty_backup,
+        classes        => classes_generic("remove_rudder_syslog_configuration_syslog");
+
+  methods:
+      "merge result classes" usebundle  => _classes_combine_three("remove_rudder_syslog_configuration_remove_files", "remove_rudder_syslog_configuration_syslogng", "remove_rudder_syslog_configuration_syslog", "remove_rudder_syslog_configuration_result");
+      "reports"              usebundle  => rudder_common_reports_generic("Common", "remove_rudder_syslog_configuration_result", "&TRACKINGKEY&", "Log system for reports", "None", "Rudder syslog configuration removed");
```
jooooooon: These are pure ncf methods. I've tested this code extensively, and it just works fine. I even tricked it into doing reporting, and the reports looked correct (they are obviously not very useful in a "remove reporting" bundle, but I figured they might come in one day...)

PR #781, opened 2015-11-24
Fixes #7460: Keep all outputs in disabled reporting mode

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7460

### DIFF
--- a/techniques/system/common/1.0/rudder_stdlib_core.st
+++ b/techniques/system/common/1.0/rudder_stdlib_core.st
@@ -123,8 +123,12 @@ bundle agent rudder_common_report(technique_name, status, identifier, component_
 bundle agent rudder_common_report_index(technique_name, status, identifier, component_name, component_key, message, index)
 {
   classes:
-    changes_only|reports_disabled::
-      "send_reports" or => {
+
+    !changes_only::
+      "log_reports" expression => "full_compliance|reports_disabled";
+
+    changes_only::
+      "log_reports" or => {
                             strcmp("${status}", "result_error") ,
                             strcmp("${status}", "result_repaired") ,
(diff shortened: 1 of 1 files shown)


techniques/system/common/1.0/rudder_stdlib_core.st
```diff
@@ -139,7 +139,7 @@ bundle agent rudder_common_report_index(technique_name, status, identifier, comp
         action    => immediate;
 
   reports:
-    full_compliance|send_reports::
+    full_compliance|reports_disabled|send_reports::
```
jooooooon: I don't like writing "print a report if reports are disabled". It just doesn't make sense.

But we do need to read these reports!

I propose instead:

```
!changes_only|(changes_only.send_reports)
```

ncharles: do we really  want to have even success in no reports ?

VinceMacBuche: I think so, since you only have the outputs  You are not able to understand what was done

At least in changes-only you can look into rudder compliance reporting

jooooooon: +1 on the change, yes. It's particularly hard to debug if you only see half of the error messages!

jooooooon: Well, messages.

VinceMacBuche:
PR updated

VinceMacBuche:
Updated the logic to set the class to be more readable and replace send_reports by log_reports which is more accurate on the intent

peckpeck:
OK Merging

PR #783, opened 2015-11-24
Fixes #7464: svcadm does not exist - so syslog is never restarted on Solaris

jooooooon:
https://www.rudder-project.org/redmine/issues/7464


peckpeck:
OK, merging

PR #784, opened 2015-11-25
Fixes #6648: \"Update\" component of common technique sometimes return unexpected status and has incomplete reporting

jooooooon:
https://www.rudder-project.org/redmine/issues/6648

### DIFF
--- a/techniques/system/common/1.0/update.st
+++ b/techniques/system/common/1.0/update.st
@@ -285,6 +285,9 @@ bundle agent update_reports
     no_update::
       "@@Common@@result_error@@&TRACKINGKEY&@@Update@@None@@${g.execRun}##${g.uuid}@#Cannot update node's policy (CFEngine promises)";
 
+    rudder_tools_updated_error::
+      "@@Common@@result_error@@&TRACKINGKEY&@@Update@@None@@${g.execRun}##${g.uuid}@#Cannot update Rudder tools last updated file";
+
     rudder_tools_update_error::
       "@@Common@@result_error@@&TRACKINGKEY&@@Update@@None@@${g.execRun}##${g.uuid}@#Cannot update Rudder tools";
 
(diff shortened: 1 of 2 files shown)


techniques/system/common/1.0/update.st
```diff
@@ -312,12 +323,9 @@ bundle agent update_reports
     config::
       "@@Common@@log_repaired@@&TRACKINGKEY&@@Update@@None@@${g.execRun}##${g.uuid}@#Node's policy (CFEngine promises) updated";
 
-    rudder_promises_generated_repaired|config|rudder_tools_updated|rudder_ncf_common_updated|rudder_ncf_local_updated|server_ok|executor_ok::
+    config|rudder_tools_updated|rudder_ncf_common_updated|rudder_ncf_local_updated|server_ok|executor_ok::
       "@@Common@@result_repaired@@&TRACKINGKEY&@@Update@@None@@${g.execRun}##${g.uuid}@#Rudder policy, tools or ncf instance were updated or CFEngine service restarted";
 
-    policy_server::
-      "@@Common@@result_success@@&TRACKINGKEY&@@Update@@None@@${g.execRun}##${g.uuid}@#Rudder policy server doesn't need to be updated";
```
jooooooon: Removing this report because it is way too simple - we just say success on either root_server or policy_server (because root_server IS a policy_server too), without checking what really happened. And if a real success (or repair) occured, this becomes "Unexpected" reports in Rudder.

jooooooon: By the way, this was introduced in 2.3 or 2.4 because there was no result_success on root_servers. However, there should be, which is why I also fixed the result_success condition above.

techniques/system/common/1.0/update.st
```diff
@@ -312,12 +323,9 @@ bundle agent update_reports
     config::
       "@@Common@@log_repaired@@&TRACKINGKEY&@@Update@@None@@${g.execRun}##${g.uuid}@#Node's policy (CFEngine promises) updated";
 
-    rudder_promises_generated_repaired|config|rudder_tools_updated|rudder_ncf_common_updated|rudder_ncf_local_updated|server_ok|executor_ok::
+    config|rudder_tools_updated|rudder_ncf_common_updated|rudder_ncf_local_updated|server_ok|executor_ok::
```
jooooooon: I removed ```rudder_promises_generated_repaired``` from the condition to print a ```result_repaired``` here because it just means that the ```rudder_promises_generated``` file was updated. This, by itself, doesn't functionnaly consitute an update. What does, is if the promises are then updated, but the ```config``` class already covers this condition.

jooooooon:
DO NOT MERGE yet - this is missing the equivalent change in initial-promises.

jooooooon:
PR updated

jooooooon:
PR updated

peckpeck:
OK, merging

PR #785, opened 2015-11-25
Fixes #7472: Promise copy is not convergent: if the copy is interupted, it will never recover (3.2 branch)

jooooooon:
https://www.rudder-project.org/redmine/issues/7472

### DIFF
--- a/techniques/system/common/1.0/update.st
+++ b/techniques/system/common/1.0/update.st
@@ -264,7 +264,7 @@ bundle agent update_action
         classes           => success("config", "no_update", "config_ok");
 
     # We copy ncf on non root server if ncf hash file changed, or if  promises have been updated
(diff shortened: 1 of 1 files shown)


techniques/system/common/1.0/update.st
```diff
@@ -264,7 +264,7 @@ bundle agent update_action
         classes           => success("config", "no_update", "config_ok");
 
     # We copy ncf on non root server if ncf hash file changed, or if  promises have been updated
-    !root_server.(rudder_promises_generated_repaired|rudder_ncf_hash_update_repaired)::
```
jooooooon: This class no longer exists in the 3.2 branch, and got here by a merge up from 3.x branches

ncharles:
ha, merge are hard :/

PR #787, opened 2015-11-26
Fixes #7142: rudder share tools : *.exe *.msi *.bat *.rpm *.vbs *.dll

jooooooon:
https://www.rudder-project.org/redmine/issues/7142

### DIFF
--- a/initial-promises/node-server/common/1.0/site.cf
+++ b/initial-promises/node-server/common/1.0/site.cf
@@ -31,7 +31,6 @@ bundle common g
       "rudder_var"                 string => "/data/rudder";
       "rudder_curl"                string => "/system/bin/curl";
       "rudder_rm"                  string => "/system/xbin/rm";
-      "rudder_tools_files_android" slist => { "cpuid-android-V1.0.sh" };
 
     !windows.!android::
       "rudder_var"                 string => "/var/rudder";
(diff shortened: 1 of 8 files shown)


initial-promises/node-server/common/1.0/site.cf
```diff
@@ -31,7 +31,6 @@ bundle common g
       "rudder_var"                 string => "/data/rudder";
       "rudder_curl"                string => "/system/bin/curl";
       "rudder_rm"                  string => "/system/xbin/rm";
-      "rudder_tools_files_android" slist => { "cpuid-android-V1.0.sh" };
```
jooooooon: These lists of files are used in ```update_action``` bundle, which is only run in failsafe mode, so they must be in ```failsafe.cf```'s copy of the ```g``` bundle.

initial-promises/node-server/failsafe.cf
```diff
@@ -192,3 +243,10 @@ body file_select exclude(name)
           file_result => "!leaf_name";
 }
 
+body file_select by_fullname(names)
+# @brief Select files that match `names`
+# @param names A regular expression
```
peckpeck: Why is this body defined twice ?

jooooooon: Once in our standard library that's included in a standard run and once in failsafe.cf because failsafe.cf doesn't use the standard libraries, only the ones in failsafe.cf itself - this is to avoid a broken stdlib breaking failsafe.cf.

peckpeck:
This PR needs a rebase

jooooooon:
PR updated

peckpeck:
OK, merging

techniques/system/common/1.0/site.st
```diff
@@ -157,29 +127,6 @@ bundle common g
...
-                                            };
-
-    SuSE::
-      "rudder_tools_files_suse"    slist => { "openvpn-2.2.1-1.x86_64.rpm",
-                                              "openvpn-2.2.1-1.i686.rpm",
-                                              "zypper-repo.tpl",
-                                              "check_zypper_version",
-                                              "checkzmd.pl",
-                                            };
-    any::
```
ncharles: an extra care should be given to class condition deletion

PR #789, opened 2015-11-26
Fixes #7176: Unexpected reports on \"cron daemon\" component of system techniques

jooooooon:
https://www.rudder-project.org/redmine/issues/7176

### DIFF
--- a/techniques/system/common/1.0/cron_setup.st
+++ b/techniques/system/common/1.0/cron_setup.st
@@ -71,6 +71,13 @@ bundle agent setup_cronjob
 
 bundle agent check_cron_daemon
 {
+  vars:
+
+    redhat|fedora::
+      "service_name" string => "crond";
+    !(redhat|fedora)::
+      "service_name" string => "cron";
+
   classes:
 
     any::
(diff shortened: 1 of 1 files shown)


techniques/system/common/1.0/cron_setup.st
```diff
@@ -80,24 +87,17 @@ bundle agent check_cron_daemon
...
-    pass3.(service_restart_cron_not_ok|service_restart_crond_not_ok)::
-      "any" usebundle => rudder_common_report("Common", "result_error", "&TRACKINGKEY&", "CRON Daemon", "None", "The CRON daemon was not running and could not be restarted");
+    pass2.restart_crond::
+      "restart_cron" usebundle => service_restart("${service_name}");
 
-    pass3.(service_restart_cron_repaired|service_restart_crond_repaired)::
-      "any" usebundle => rudder_common_report("Common", "result_repaired", "&TRACKINGKEY&", "CRON Daemon", "None", "The CRON daemon has been successfully restarted");
+    # Fake success result, since we don't actually call service_restart unless it's needed
+    pass2.!restart_crond::
+      "any" usebundle => _classes_success("service_restart_${service_name}");
```
peckpeck: Shouldn't we do a result_na instead when there is no crond running ?

jooooooon: We do - it's just a few lines down from this one.

This case is so we have a success report when crond *is* running, but doesn't need restarting.

peckpeck: restart_crond is only set when crond is not running

jooooooon: Yes, exactly. So when it's *not* set we declare success. But actually, we should add in ```.!(android|windows|aix)``` to avoid duplicate report with the NA below!

jooooooon:
PR updated

techniques/system/common/1.0/cron_setup.st
peckpeck: Since restart_crond is never set when we are under thos os, this is redudant.
But the original question was about doing a result_na instead of a result_sucess in case there is no crond runnning.

techniques/system/common/1.0/cron_setup.st
```diff
@@ -80,24 +87,17 @@ bundle agent check_cron_daemon
...
-    pass3.(service_restart_cron_not_ok|service_restart_crond_not_ok)::
-      "any" usebundle => rudder_common_report("Common", "result_error", "&TRACKINGKEY&", "CRON Daemon", "None", "The CRON daemon was not running and could not be restarted");
+    pass2.restart_crond::
+      "restart_cron" usebundle => service_restart("${service_name}");
 
-    pass3.(service_restart_cron_repaired|service_restart_crond_repaired)::
-      "any" usebundle => rudder_common_report("Common", "result_repaired", "&TRACKINGKEY&", "CRON Daemon", "None", "The CRON daemon has been successfully restarted");
+    # Fake success result, since we don't actually call service_restart unless it's needed
+    pass2.!(android|windows|aix).!restart_crond::
+      "any" usebundle => _classes_success("service_restart_${service_name}");
```
jooooooon: The class condition here is ```!restart_crond``` - note the *NOT*

This is by default true everywhere, and therefore on ```android|windows|aix```, which is why I have listed them here, negated.

If crond is not running, then the restart_class will be set, and this condition will not be true, so it's irrelevant.

peckpeck:
OK Merging

PR #790, opened 2015-11-26
Fixes #7389: Syslog fails to restart due bad line in syslog-ng.conf

jooooooon:
https://www.rudder-project.org/redmine/issues/7389


peckpeck:
Good, merging

PR #791, opened 2015-11-27
Fixes #7476: Initial promises try to copy files and set group to "roo…

ncharles:
…t" rather than 0


jooooooon:
This is a good start but there are some missing:
- initial-promises/node-server/common/1.0/process_matching.cf:62
- initial-promises/node-server/server-roles/1.0/network-check.cf:45
- initial-promises/node-server/server-roles/1.0/password-check.cf:74

There are also 2 more missing in update.cf.

jooooooon:
PR updated

jooooooon:
Replaced by #794

PR #794, opened 2015-11-27
Fixes #7476: Initial promises try to copy files and set group to \"root\" rather than 0

jooooooon:
https://www.rudder-project.org/redmine/issues/7476


ncharles:
oh, but the ticket name is not ok then ...

jooooooon:
> oh, but the ticket name is not ok then ...

Fixed.

PR #796, opened 2015-11-27
As per bash man:

coredumb:
/etc/profile
  The systemwide initialization file, executed for login shells

Problem is that cf-agent seems to not fully have environment variables as not being in a login shell.
Not sure how this translate on a plain /bin/sh - not symlinked to /bin/bash - nor if it's standardized across distributions.

### DIFF
--- a/techniques/system/common/1.0/environment-variables.st
+++ b/techniques/system/common/1.0/environment-variables.st
@@ -27,7 +27,8 @@ bundle agent get_environment_variables
     # will be available using the syntax ${node.env[ENV_NAME]}
 
     !windows::
-      "script" string => "#! /bin/sh
+      "script" string => "#!/bin/sh
+source /etc/profile
 env | sed 's/=/]=/' |sed 's/^/=node.env[/'";
 
     !windows.linux::


fanf:
@amousset : could take an eye on that one? Open a ticket if needed, etc.

techniques/system/common/1.0/environment-variables.st
```diff
@@ -27,7 +27,8 @@ bundle agent get_environment_variables
     # will be available using the syntax ${node.env[ENV_NAME]}
 
     !windows::
-      "script" string => "#! /bin/sh
+      "script" string => "#!/bin/sh
+source /etc/profile
```
jooooooon: This should be "." not source, because "source" is a bashism.

Also, maybe we should check that /etc/profile actually exists first, like ```[ -f /etc/profile ] && . /etc/profile```

coredumb: Damn you're right.

amousset:
Opened #921 with the fixes, closing this one.

PR #797, opened 2015-11-27
Fixes #7496: rudder-techniques test script \"technique-files\" doesn't list promises.cf and failsafe.cf in initial-promises

jooooooon:
https://www.rudder-project.org/redmine/issues/7496

### DIFF
--- a/scripts/technique-files
+++ b/scripts/technique-files
@@ -17,12 +17,13 @@ set -euo pipefail
 LAST=0
 HELP=0
 DIR_ONLY=0
+INCLUDE_INITIAL=0
 PATTERN=""
 FILTER=""
 PLACEHOLDER=1
 
 # parameter parsing
-while getopts "hldf:s:p" opt
+while getopts "hldif:s:p" opt
 do
   case ${opt} in
     h)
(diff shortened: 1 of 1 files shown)


scripts/technique-files
```diff
@@ -96,6 +101,12 @@ find_directories() {
       fi
 }
 
+# find extra 'special' files in initial-promises root (promises.cf, failsafe.cf, ...)
+find_initial() {
+  if [ ${INCLUDE_INITIAL} -eq 1 ] && [ -d ${directory}/initial-promises/node-server ]; then
+    find "${directory}/initial-promises/node-server/" -maxdepth 1 -not -wholename "*/.git/*" -type f
```
peckpeck: since directory can be any directory, including initial-promises itself, I'd prefer something like
node_server=$(find "${directory}" -type d -name node-server)
and use this variable instead of the literal ${directory}/initial-promises/node-server

jooooooon:
PR updated

jooooooon:
PR updated

peckpeck:
OK, merging

PR #798, opened 2015-11-28
Fixes #7499: Use of nonexistant class cfengine_community makes failsafe.cf not copy CFEngine binaries to /var/rudder

jooooooon:
https://www.rudder-project.org/redmine/issues/7499


jooooooon:
PR updated

PR #799, opened 2015-11-28
Fixes #7004: Invalid inventory files only partially moved/renamed to the fail folder

jooooooon:
https://www.rudder-project.org/redmine/issues/7004

### DIFF
--- a/tools/send-clean.sh
+++ b/tools/send-clean.sh
@@ -87,7 +87,8 @@ elif [ ${SEND_COMMAND_RET} -ne 0 -o "${HTTP_CODE}" -ge 400 ]
 then
   # The server refused our inventory (or there was a problem)
   echo "ERROR: Failed to send inventory ${FILENAME}, putting it in the failed directory"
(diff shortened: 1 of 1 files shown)


tools/send-clean.sh
```diff
@@ -87,7 +87,8 @@ elif [ ${SEND_COMMAND_RET} -ne 0 -o "${HTTP_CODE}" -ge 400 ]
 then
   # The server refused our inventory (or there was a problem)
   echo "ERROR: Failed to send inventory ${FILENAME}, putting it in the failed directory"
-  mv "${FILENAME}" "${FAILEDDIR}/${BASENAME}-$(date --rfc-3339=date)"
+  mv "${FILENAME}" "${FAILEDDIR}/`basename ${FILENAME}`-$(date --rfc-3339=date)"
```
jooooooon: The reason for this change is that ```FILENAME``` gets updated when the ```.gz``` suffix is removed, but ```BASENAME``` does not. And it is preferable to store the failed inventory *without* the ```.gz``` suffix, to make it easier to analyse afterwards (ie, don't try and decompress a plain text file!)

peckpeck:
OK, merging

PR #800, opened 2015-11-28
Fixes #6082: Make CFEngine binaries usable by everyone (mode 755)

jooooooon:
https://www.rudder-project.org/redmine/issues/6082


jooooooon:
PR updated

ncharles:
Why targeting only branch 3.2 ?
Default cfengine installs have 755 or all binaries

jooooooon:
> Why targeting only branch 3.2 ?
> Default cfengine installs have 755 or all binaries

This is a significant change. If users have audited a Rudder installation, and care about file perms, we should not make a change like this in a patch version.

ncharles:
Oh, ok then

PR #804, opened 2015-11-30
Fixes #7504: Add test for rudder-techniques to ensure removed deprecated techniques don't come back in an erroneous merge

jooooooon:
https://www.rudder-project.org/redmine/issues/7504


scripts/check-techniques.sh
```diff
@@ -146,3 +146,11 @@ do
...
...shConfiguration/3.0 systemSettings/remoteAccess/sshKeyDistribution/1.0 systemSettings/remoteAccess/sshKeyDistribution/2.0 systemSettings/security/fileAlterationMonitoring/1.0 systemSettings/systemManagement/cronManagement/1.0 systemSettings/systemManagement/cronManagement/2.0 systemSettings/systemManagement/fstabConfiguration/1.0 systemSettings/systemManagement/fstabConfiguration/2.0 systemSettings/systemManagement/fstabConfiguration/3.0 systemSettings/systemManagement/motdConfiguration/1.0 systemSettings/systemManagement/motdConfiguration/2.0 systemSettings/systemManagement/motdConfiguration/3.0 systemSettings/userManagement/groupManagement/1.0 systemSettings/userManagement/groupManagement/2.0 systemSettings/userManagement/groupManagement/3.0 systemSettings/userManagement/groupManagement/3.1 systemSettings/userManagement/groupManagement/4.0 systemSettings/userManagement/sudoParameters/1.0 systemSettings/userManagement/sudoParameters/2.0 systemSettings/userManagement/userManagement/1.0 systemSettings/userManagement/userManagement/2.0 systemSettings/userManagement/userManagement/3.0 systemSettings/userManagement/userManagement/4.0 systemSettings/userManagement/userManagement/5.0; do
```
peckpeck: How will you make sure that this list is kept up to date ?

jooooooon: No idea :(

This is currently taken from the git commit that removed the techniques. I guess the best I can think of is that us PR reviewers make sure any future PRs that remove technique versions also update this script...

peckpeck: We shouldn't rely on people's exceptional memory to check something like this.
Maybe an equivalent  test in 2.11 that test technique presence.
Then the test has to be updated on each modification and web cannot forget it.

ncharles: Here, it's for technique that we removed - checking the presence in lower branches won't help, as they are expected to be there
Deletion can only happen (roughly) during the branching, and the list won't change for all the life of the branch

peckpeck: The process would be :
* we add a presence check in 2.11
* the check is upmerged
* we remove a technique in 3.0
* -> the presence test fails
* we know we need to change the test so we change the presence test into absence test
* a technique is wrongly upmerge
* -> the test alerts us

jooooooon: Right, thanks for the clarification, I understand a lot better now!

If we do that, we should probably add another test to detect any unmanaged technique versions, so that when we add new versions we don't forget to add them too.

However, this is beginning to sound like a lot of work...

peckpeck: technique_present="..."
technique_absent="..."
for i in present; do [ -d $i ] || error; done
for i in absent; do [ -d $i ] && error; done
for i in $(technique-files -d techniques); do OK=0; for j in present absent; do [ $i =$j ] && OK=1; done [ $OK = 0 ] && error; done

I'm just missing the list and the error function

jooooooon: Aha, thanks :))

But I didn't mean complicated to implement, I mean complicated to maintain: every time we add a technique version, we're going to forget to update this list, and therefore the tests will fail, we'll have to open another ticket, make another commit, etc etc etc

peckpeck: Yes but that's better than forgetting to add a test and never knowing it since it won't fail.

jooooooon: OK, your call.

jooooooon:
I've adapted this and proposed an implementation of the approach you describe here in #841

PR #809, opened 2015-11-30
Fixes #7508: The rudder-agent cron on AIX uses if then, which makes some security test fails

jooooooon:
https://www.rudder-project.org/redmine/issues/7508


ncharles:
didn't you forget initial promises ?

jooooooon:
PR updated

jooooooon:
> didn't you forget initial promises ?

Good catch. For some reason the same code is in a different file... caught me out :/

ncharles:
yeah, i've noticed the cron part is really different on both block of code

PR #811, opened 2015-11-30
Fixes #7518: Promise types out of order in initial promises' promises.cf

jooooooon:
https://www.rudder-project.org/redmine/issues/7518


peckpeck:
OK merging

PR #815, opened 2015-12-11
Fixes #7599: Modify commons techniques to add rudder-groups.st and related info

fanf:
https://www.rudder-project.org/redmine/issues/7599

### DIFF
--- a/techniques/system/common/1.0/rudder-groups.st
+++ b/techniques/system/common/1.0/rudder-groups.st
@@ -1,0 +1,43 @@
+#####################################################################################
+# Copyright 2015 Normation SAS
+#####################################################################################
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, Version 3.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# This file contains all the groups where the node belongs to. The groups are available on two ways:
+# 1/ two classes are added for each groups: 
+#    - group_UUID with UUID replaced by the group unique identifier
+#    - group_NAME with NAME replaced by the transliteration of the group name to lower-case ascii, and the 
+#      canonification of the result
+# 2/ an array "by_uuid" of UUID -> ORIG_NAME, available as a global variable (for example for processing
+#    all node groups to use in templates)
+#  
+#  For example, if node belongs to group: (id: 64f85ba8-39c7-418a-a099-24c2c2909dfd ; name: "Serveurs pré-prod")
+#  we will have the following classes:
+#  - group_64f85ba8_39c7_418a_a099_24c2c2909dfd
+#  - group_serveurs_pre_prod
+#  and vars:
+#  - by_uuid["64f85ba8-39c7-418a-a099-24c2c2909dfd"] = "Serveurs pré-prod"
+# More over, each value of the array will have an attribute "meta": { "inventory", "attribute_name=rudder_groups" }
+#
+bundle common rudder_groups {
+
+  vars:
+      &RUDDER_NODE_GROUPS_VARS&
+
+  classes:
+      &RUDDER_NODE_GROUPS_CLASSES&
+
+}
--- a/techniques/system/common/1.0/metadata.xml
+++ b/techniques/system/common/1.0/metadata.xml
@@ -66,6 +66,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     </TML>
     <TML name="environment-variables"/>
     <TML name="restart-services"/>
+    <TML name="rudder-groups"/>
   </TMLS>
 
   <SYSTEMVARS>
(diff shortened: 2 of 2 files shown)


techniques/system/common/1.0/rudder-groups.st
```diff
@@ -0,0 +1,42 @@
+#####################################################################################
+# Copyright 2013 Normation SAS
```
jooooooon: 2015

techniques/system/common/1.0/rudder-groups.st
```diff
@@ -0,0 +1,42 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# This file contains all the groups where the node belongs to. The groups are available on two ways:
+# 1/ two classes are added for each groups: 
+#    - group_UUID with UUID replaced by the group unique identifier
+#    - group_NAME with NAME replaced by the transliteration of the group name to ascci, and the 
+#      normalisation of the result
```
jooooooon: The correct term in CFEngine is "canonification"

techniques/system/common/1.0/rudder-groups.st
```diff
@@ -0,0 +1,42 @@
...
+# 2/ an array "by_uuid" of UUID -> ORIG_NAME, available as a global variable (for example for processing
+#    all node groups to use in templates)
+#  
+#  For example, if node belongs to group: (id: 64f85ba8-39c7-418a-a099-24c2c2909dfd ; name: "Serveurs pré-prod")
+#  we will have the following classes:
+#  - group_64f85ba8_39c7_418a_a099_24c2c2909dfd
+#  - group_Serveur_pre_prod
+#  and var:
```
jooooooon: s/var:/vars:/

techniques/system/common/1.0/rudder-groups.st
```diff
@@ -0,0 +1,42 @@
...
+#    - group_UUID with UUID replaced by the group unique identifier
+#    - group_NAME with NAME replaced by the transliteration of the group name to ascci, and the 
+#      normalisation of the result
+# 2/ an array "by_uuid" of UUID -> ORIG_NAME, available as a global variable (for example for processing
+#    all node groups to use in templates)
+#  
+#  For example, if node belongs to group: (id: 64f85ba8-39c7-418a-a099-24c2c2909dfd ; name: "Serveurs pré-prod")
+#  we will have the following classes:
+#  - group_64f85ba8_39c7_418a_a099_24c2c2909dfd
+#  - group_Serveur_pre_prod
```
jooooooon: According to the example above this is missing a "s" (group_Serveurs_pre_prod).

CFEngine class names are case-sensitive... Maybe we should just lower-case the name aswell to avoid confusion?

techniques/system/common/1.0/rudder-groups.st
```diff
@@ -0,0 +1,42 @@
...
+#  and var:
+#  - by_uuid["64f85ba8-39c7-418a-a099-24c2c2909dfd"] string => "Serveurs pré-prod"
+# 
+bundle common rudder_groups {
+
+  vars:
+    &RUDDER_NODE_GROUPS_VARS&
+
+  classes:
+    any::
```
jooooooon: This is the default. Why include it here?

techniques/system/common/1.0/rudder-groups.st
fanf: Right for the s. 
For the lower case, I was wondering to. As Rudder forbid groups with the same name (case insensitive), I think i would be more consistent.

techniques/system/common/1.0/rudder-groups.st
fanf: I took example on other template and thought it was the convention.

techniques/system/common/1.0/rudder-groups.st
```diff
@@ -0,0 +1,42 @@
...
+#      normalisation of the result
+# 2/ an array "by_uuid" of UUID -> ORIG_NAME, available as a global variable (for example for processing
+#    all node groups to use in templates)
+#  
+#  For example, if node belongs to group: (id: 64f85ba8-39c7-418a-a099-24c2c2909dfd ; name: "Serveurs pré-prod")
+#  we will have the following classes:
+#  - group_64f85ba8_39c7_418a_a099_24c2c2909dfd
+#  - group_Serveur_pre_prod
+#  and var:
+#  - by_uuid["64f85ba8-39c7-418a-a099-24c2c2909dfd"] string => "Serveurs pré-prod"
```
ncharles: missing ; at the end
and could you add a  meta    => { "inventory", "attribute_name=rudder_groups" }; ?

fanf: That was not suppose to be a valid cfengine line, no more than the classes are. 
The actual result is in the test cases here: https://github.com/Normation/rudder/pull/995/files#diff-0058ac7299eebb012e0f765baf965529R37

fanf: I can nonetheless change the wording to be valid cfengine in both vars and classes examples

techniques/system/common/1.0/rudder-groups.st
jooooooon: No, it's not necessary.

techniques/system/common/1.0/rudder-groups.st
jooooooon: OK, let's do it. Don't forget to update the comments above to explain that too.

fanf:
pr updated

techniques/system/common/1.0/rudder-groups.st
```diff
@@ -0,0 +1,43 @@
...
+#  - by_uuid["64f85ba8-39c7-418a-a099-24c2c2909dfd"] = "Serveurs pré-prod"
+# More over, each value of the array will have an attribute "meta": { "inventory", "attribute_name=rudder_groups" }
+#
+bundle common rudder_groups {
+
+  vars:
+    &RUDDER_NODE_GROUPS_VARS&
+
+  classes:
+      &RUDDER_NODE_GROUPS_CLASSES&
```
ncharles: why not the same identation between both ?
i think the previous one miss 2 spaces:
2 for vars:
2 for (optionnal) class condition
2 for the promisers

fanf: because I didn't remove 2 spaces when deleting any:: :)

jooooooon: Nor shuold you. But you should add 2 spaces in front of ```&RUDDER_NODE_GROUPS_VARS&```

fanf: ok

fanf:
pr updated

PR #819, opened 2015-12-16
Fixes #7621: cf-promises can hang for about 30 seconds looking up default server name \"rudder\"

jooooooon:
https://www.rudder-project.org/redmine/issues/7621


techniques/system/common/1.0/cf-served.st
```diff
@@ -81,7 +81,7 @@ bundle server access_rules
   &endif&
 
 &endif&
-    policy_server_set::
+
```
jooooooon: Using this class is no longer necessary, as the new implementation will *always* have a policy_server defined - which may be "rudder", but it will have one.

techniques/system/common/1.0/update.st
```diff
@@ -99,14 +99,11 @@ bundle common server_info
...
-    policy_server_set::
-      "cfserved" string => readfile("${policy_server_file}", 2048);
-    !policy_server_set::
-      "cfserved" string => "rudder";
-
-  classes:
-      "policy_server_set" expression => fileexists("${server_info.policy_server_file}");
+      # Use "rudder" as default policy_server if no policy_server.dat exists
+      "cfserved" string => ifelse(fileexists("${server_info.policy_server_file}"),
+                                  readfile("${policy_server_file}", 2048),
```
ncharles: i'm not sure you can safely use functions within ifelse - at least i'm pretty sure that it will fail with agent using cfengine 3.5

jooooooon: It works on our 3.6.5 agent.

Not sure about 3.5... you're right though, we should test.

techniques/system/common/1.0/cf-served.st
```diff
@@ -98,15 +98,12 @@ bundle common def
...
         comment => "Path to file containing address to policy server";
 
-   # Use localhost as default policy_server if no policy_server.dat exists
-    policy_server_set::
+      # Use "rudder" as default policy_server if no policy_server.dat exists
       "policy_server"
-        string  => readfile("${policy_server_file}", 2048),
+        string  => ifelse(fileexists("${def.policy_server_file}"),
+                          readfile("${policy_server_file}", 2048),
+                          "rudder"),
```
ncharles: this does not work correctly on Rudder Agent 2.10
If the file is not there, the output of the cf-serverd is
2016-01-12T11:43:47+0100  verbose: Path '/var/rudder/cfengine-community/bin/cf-agent' (encrypt=0)
2016-01-12T11:43:47+0100  verbose: Admit: '�P	`kQ	' root=
2016-01-12T11:43:47+0100  verbose: Admit: '��O	��K	' root=

....

2016-01-12T11:43:47+0100  verbose: IP '${def.policy_server}'
2016-01-12T11:43:47+0100  verbose: IP '�0N	�nQ	'
2016-01-12T11:43:47+0100  verbose: IP 'PK	��R	'

ncharles: however it does work with rudder 2.11 and more

jooooooon: So since 2.10 is not supported, that should be OK then?

ncharles: Officially, 2.11 should work with 2.10 agent

jooooooon: Ah, right. Thanks for clarifying. I'll look into this then.

peckpeck:
@jooooooon we do not support agent 2.10 from server 3.1 anymore.
This means we can add anything that use cfengine 3.6 syntax in 3.1.

Can you retarget this PR to 3.1 ?

jooooooon:
PR replaced by https://github.com/Normation/rudder-techniques/pull/918

PR #824, opened 2015-12-17
Fixes #7598: Adding new features and timezones

matya:
- NTP Servers
    * can hold multiple server in one line, separated by commas
    * final list is stripped of duplicates

- Timezones
    * Adding all standard timezones
    * Allowing textbox (useful for variable usage)

- RTC Clock
    * Allow to be set to UTC or Localtime or not change
    * Is only set on systems actually having an RTC clock

- Xen Time
    * New feature: Allow setting of Xen Hypervisor time via Dom0


ncharles:
ha, commit is too big for me to comment on it
So, the bundle list in metadata is not correct ( https://github.com/matya/rudder-techniques/blob/matya-7598-add-features-to-timesettings/techniques/systemSettings/misc/clockConfiguration/4.0/metadata.xml#L33 ). You should have two entries NAME, one for each bundle

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
```diff
@@ -0,0 +1,462 @@
...
+# Change the server configuration, and restart the daemon
+# if the configuration changed
+# Caution : a missing feature in cfe prevents from adding ,0x1 at the end of the domain name
+# so it's only safe to use with ips for now
+
+bundle agent check_clock_configuration
+{
+
+  vars:
+      "ntpServers_all" slist => {&CLOCK_NTPSERVERS: { splitstring("&it&", " *, *", 100)};separator=", "&};
```
ncharles: i am amazed that it does work !

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
```diff
@@ -0,0 +1,462 @@
...
+      "target_windows_timezone" string => "${tz_variables.tz_win[${target_timezone}]}";
+      "current_windows_timezone" string => execresult("\"${sys.winsysdir}\tzutil.exe\" /g", "useshell");
+
+    linux::
+      "linux_timezone"   string => "${tz_variables.tz_linux[${target_timezone}]}";
+
+  classes:
+
+      "clock_timezone_nochange"  expression => strcmp("&CLOCK_TIMEZONE&","dontchange");
+      "clock_timezone_valid" expression => isvariable("tz_variables.tz_linux[${target_timezone}]");
```
ncharles: here, you check only the validity on Linux, not windows
you should have a case linux:: 
"clock_timezone_valid" expression => isvariable("tz_variables.tz_linux[${target_timezone}]");
and
windows::
"clock_timezone_valid" expression => isvariable("tz_variables.tz_win[${target_timezone}]");

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
```diff
@@ -0,0 +1,462 @@
...
+      "clock_timezone_noedit" expression => "clock_timezone_nochange|!clock_timezone_valid";
+
+      "clock_hwclock_sync" expression => strcmp("&CLOCK_HWSYNC_ENABLE&","true");
+      "clock_date_s_now" expression => strcmp("&CLOCK_DATE_S_NOW_ENABLE&","true");
+
+      "clock_hwclock_has_rtc" expression => fileexists("/dev/rtc");
+
+      "ntp_config_file_exists" expression => fileexists("/etc/ntp.conf");
+
+    clock_timezone_valid.windows::
```
ncharles: cf previous remark, it is not possible to have clock_timezone_valid and windows at the same time with this implementation

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
```diff
@@ -0,0 +1,462 @@
...
+        edit_defaults => noempty_backup,
+        classes => kept_if_else("redhattz_kept", "repaired_redhattz", "not_repaired_redhattz");
+
+    (SuSE.!clock_timezone_noedit).clock_timezone_valid::
+      "/etc/sysconfig/clock"
+        edit_line => EditSuSETimezone("${linux_timezone}", "${clock_hwclock_rtc}"),
+        edit_defaults => noempty_backup,
+        classes => kept_if_else("susetz_kept", "repaired_susetz", "not_repaired_susetz");
+
+    ## TODO: Handle UTC/Localtime on Debian
```
ncharles: Ha. Do you have any idea on how it should be done ?

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
```diff
@@ -0,0 +1,462 @@
...
+    pass3.ntp_hwclock_sync_error::
+      "any" usebundle => rudder_common_report("ntpConfiguration", "result_error", "&TRACKINGKEY&", "Hardware clock (RTC)", "None", "The hardware clock could not be synchronized with the NTP time");
+
+    pass3.clock_hwclock_sync.!clock_hwclock_has_rtc::
+      "any" usebundle => rudder_common_report("ntpConfiguration", "result_na", "&TRACKINGKEY&", "Hardware clock (RTC)", "None", "No hardware clock present on /dev/rtc");
+
+    pass3.windows.!clock_hwclock_rtc2utc::
+      "any" usebundle => rudder_common_report("ntpConfiguration", "result_success", "&TRACKINGKEY&", "Hardware clock (RTC)", "None", "The hardware clock is automatically synchronized with the NTP time on Windows");
+
+    pass3.rtc_regkept::
```
ncharles: you should never use only kept, the risk is too great to have a pass defining a repaired, and another one defining kept, and having wrong reporting
So it should be
pass3.rtc_regkept.!rtc_regfail.!rtc_regset

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
```diff
@@ -0,0 +1,462 @@
...
+      "any" usebundle => rudder_common_report("ntpConfiguration", "result_success", "&TRACKINGKEY&", "Hardware clock (RTC)", "None", "RTC set to UTC");
+
```
ncharles: you miss a repaired report here

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
```diff
@@ -0,0 +1,462 @@
...
+    pass3.!clock_hwclock_sync::
+      "any" usebundle => rudder_common_report("ntpConfiguration", "result_success", "&TRACKINGKEY&", "Hardware clock (RTC)", "None", "No synchronization with the hardware clock was requested");
+
+    pass3.!clock_date_s_now::
+      "any" usebundle => rudder_common_report("ntpConfiguration", "result_na", "&TRACKINGKEY&", "Xen Time", "None", "Setting Time is disabled");
+
+    pass3.clock_date_s_now.!xen_dom0::
+      "any" usebundle => rudder_common_report("ntpConfiguration", "result_na", "&TRACKINGKEY&", "Xen Time", "None", "Not on a Xen Dom0");
+
+    pass3.date_s_set::
```
ncharles: it should be pass3.date_s_set.!date_s_failed

ncharles:
this is one huge PR ! thank you very much for it, I've made some remark about the some minor points, that should be adressed
I have a question also: what happens if we set CLOCK_HWSYNC_RTC and a timezone on windows ? Doesn't it conflict (it would run tzutil even if not necessary, isn't it ?)

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
matya: Just some cascaded splitting logic

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
matya: Yeah, I was thinking about that. My idea was that the whole windows/linux part was not coming from the user but is generated and provided by the system, so of one is present, so is the other. But you are right, just to have all edges covered I'll modify it it check it that way to catch all possible issues.

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
matya: You can. It's defined even if you check against the Linux on Windows, since the keys are identical for Windows/linux, so checking one also checks the other. But I'll change the previous noted issue so this will have a straight forward logic

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
matya: Depends on the version. I did some googling:
https://www.kariliq.nl/misc/debian-clock.html
But for my approach missing this is not worse then the previous version of the technique, where it was also not handled... This would require some research and testing for debian/Ubuntu which I had no time for...

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
ncharles: Ooooohhhhhh !
A comment then would be most welcomed, but I think that having two cases, one for linux and one for windows would be more future proof (and user proof :) )

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
matya: Well, I just mimiced the same as ntp_regset, and that was designed the same way this is, so if you say this, then "pass3.ntp_regset::" should also be modified around line 185...

matya:
@ncharles Some short explanation what's going on...
RTC time and the Timezone of Windows are two different settings, however, they influence each other.

1. **RTC** is the **Real-Time Clock**, that ticks on every motherboard, and sets the date/time on boot, before NTP can kick in. Useful if you have Databases starting before NTP, or for your log files not having a timejump from t=0 to t=<now>... 
2. **UTC** is the "**Coordinated Universal Time**", which has no Timezone, no Leapsecond, and should actually be used to feed the RTC... 
... however, by default Windows stores Local time in the RTC, and this depends on the timezone set.  When you enable UTC for RTC and set a Timezone, windows should set the RTC always to UTC and the displayed time would depend on the Timezone.

I have currently no way of testing this additional code on a windows-client with NOVA, so the code currently is only "theoretical". After some google-work I found, that this registry-setting seem fail to work starting from Win8+: [Hardware Clock in UTC](https://social.technet.microsoft.com/Forums/windows/en-US/84dd824b-dd35-4d71-8d92-dbc29c78a2a4/hardware-clock-in-utc?forum=w8itproinstall)

This was introduced because of the well-known [PEBKAC-Problem](https://en.wikipedia.org/wiki/User_error), which hit the american people with their time difference even more in the early days, and now they are just keeping it for "historical reasons": [Why does Windows keep your BIOS clock on local time?](https://blogs.msdn.microsoft.com/oldnewthing/20040902-00/?p=37983)

I could throw it out, then the feature-set would stay as version 3.0, so no worsening on that, or this would need testing and would work with Win7 if I bet correctly... The call is yours which way to go ;-)

techniques/systemSettings/misc/clockConfiguration/4.0/clockConfiguration.st
matya: I am throwing out this "date -s now" code as it's too specific for our problem, and is not ever 100% correct in some corner-cases, and would only confuse people, so this will be gone in the next version of the PR.

matya:
@ncharles After our discussion on IRC I'll close this to split up the portions in different tickets and PR's

PR #825, opened 2015-12-18
Fixes #7643: Technique copyGitFile version 2.0 can send success AND error reports on post-hook execution and lead to unexpected reports (3.

jooooooon:
https://www.rudder-project.org/redmine/issues/7643


peckpeck:
OK merging

PR #827, opened 2015-12-22
Fixes #7680: check-rsyslog-version is missing on the nodes

amousset:
https://www.rudder-project.org/redmine/issues/7680


techniques/system/common/1.0/site.st
```diff
@@ -145,7 +145,7 @@ bundle common g
       "rudder_tools_files_unix"    slist => { "send-clean.sh",
                                               "cpuid-linux-V1.0.sh",
                                               "vmware_info.sh",
-                                              "check_rsyslog_version",
+                                              "check-rsyslog-version",
```
amousset: The name of the file was wrong since at least 2.11 but the selection according to the platform is actually enforced only in 3.2 (https://www.rudder-project.org/redmine/issues/7142).

PR #828, opened 2015-12-22
Fixes #7644: Permissions of rudder tools are not preserved on relays

amousset:
https://www.rudder-project.org/redmine/issues/7644

### DIFF
--- a/techniques/system/common/1.0/update.st
+++ b/techniques/system/common/1.0/update.st
@@ -33,7 +33,7 @@ body copy_from remote(server, path)
         trustkey   => "true";
         source     => "${path}";
         compare    => "digest";
-        preserve   => "false"; #preserver permissions
+        preserve   => "false"; #do not preserve permissions
         verify     => "true";
         purge      => "true";
     community_edition::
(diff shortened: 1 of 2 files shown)


techniques/system/common/1.0/update.st
```diff
@@ -320,7 +320,7 @@ bundle agent update_action
         comment           => "Check if we need to update the tools";
 
     # We copy only if we have new tools based on the timestamp
-    new_tools_available::
+    !policy_server.new_tools_available::
```
amousset: In case there is already a rudder_tools_updated file.

ncharles:
are you sure the issue is only on 3.2 ?

amousset:
The problem in the code is there in 3.0 (and 2.11) too, the difference is somehow in the initial copy which preserved the permissions before 3.2. If a file in Rudder tools is then changed on the server, it triggers the update.cf copy which preserves permissions, so we never saw the permission problem happen.
Retargeting to 2.11.

amousset:
Replaced by #828.

PR #829, opened 2015-12-23
Fixes #7644: Permissions of rudder tools are not preserved on relays

amousset:
https://www.rudder-project.org/redmine/issues/7644


techniques/system/common/1.0/update.st
```diff
@@ -223,7 +223,7 @@ bundle agent update_action
         touch      => "true";
 
     # Copy the tools only if the file rudder_tools_updated is not up to date
-    any::
+    !policy_server::
```
jooooooon: This change needs to be reflected in the reporting section below - currently, there will be no success report on root and relay servers!

PR #830, opened 2015-12-24
Fixes #7690: Use classic protocol in Rudder 3.2

amousset:
https://www.rudder-project.org/redmine/issues/7690

### DIFF
--- a/initial-promises/node-server/failsafe.cf
+++ b/initial-promises/node-server/failsafe.cf
@@ -28,6 +28,7 @@ body common control
         inputs => { "common/1.0/update.cf" };
         output_prefix => "rudder";
 
+        protocol_version   => "classic";
 }
 
 body agent control {
--- a/initial-promises/node-server/promises.cf
+++ b/initial-promises/node-server/promises.cf
@@ -50,6 +50,8 @@ body common control
 
     any::
         output_prefix => "rudder";
+        
+        protocol_version   => "classic";
 
         inputs => {
   @{va.inputs_list}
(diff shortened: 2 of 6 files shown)


peckpeck:
OK, good, but why not in 2.11 ? especially for system techniques

peckpeck:
Lets backport it later

ncharles:
protocl_version was introduced in 3.6, so won't work on 3.5
We need to keep the compatibility with previous version of Rudder in Rudder <3.1 (unless stated otherwise)
So only Rudder 3.2 will have this fix

peckpeck:
2.11 if cfengine 3.6
And older server will give their promises to newer agents from their system techniques, so newer agent won't be able to talk to older server.

ncharles:
@peckpeck no, server 2.11 may still server agent 2.10, and so it will break.

ncharles:
and if you use agent 3.2 with server 2.11, it will use classic protocol, same for agent 3.1 with server 3.2

peckpeck:
It detects the server version before choosing the protocol ?

ncharles:
CFEngine 3.6 uses classical by defaut.
If any of the two use classical, the protocol used is classical

peckpeck:
ok

ncharles:
this is why agent 3.1 + server 3.2 works with NAT system, but without this patch, agent 3.2 and server 3.2 does not

amousset:
yes but agent 3.2 with any server from 2.11 will use TLS as it is supported by the server.

PR #831, opened 2015-12-28
Fixes #7678: Non quiet curl calls on root and relay servers

jooooooon:
https://www.rudder-project.org/redmine/issues/7678

### DIFF
--- a/techniques/system/server-roles/1.0/password-check.st
+++ b/techniques/system/server-roles/1.0/password-check.st
@@ -271,8 +271,7 @@ bundle agent root_password_check_dav
 
   classes:
 
-      "dav_cant_connect" not => returnszero("/usr/bin/curl -s -f -u ${g.davuser}:${g.davpw} -T /opt/rudder/etc/uuid.hive http://localhost/inventory-updates/uuid.hive","noshell");
-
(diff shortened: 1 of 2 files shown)


jooooooon:
PR updated

techniques/system/server-roles/1.0/password-check.st
```diff
@@ -271,8 +271,7 @@ bundle agent root_password_check_dav
 
   classes:
 
-      "dav_cant_connect" not => returnszero("/usr/bin/curl -s -f -u ${g.davuser}:${g.davpw} -T /opt/rudder/etc/uuid.hive http://localhost/inventory-updates/uuid.hive","noshell");
-
+      "dav_cant_connect" not => returnszero("/usr/bin/curl -s -f -o /dev/null -u ${g.davuser}:${g.davpw} -T /opt/rudder/etc/uuid.hive http://localhost/inventory-updates/uuid.hive","noshell");
```
jooooooon: We need to send the output to /dev/null here because CFEngine sends stdout from returnszero commands to it's own stdout, making the output messy

jooooooon:
PR updated

peckpeck:
OK, merging

PR #832, opened 2015-12-28
Fixes #7697: Fix permissions if the rudder tools in the repo

amousset:
https://www.rudder-project.org/redmine/issues/7697


Normation-Rudder-Redmine-readonly:
OK, merging

PR #835, opened 2015-12-31
Fixes #7714: Broken rudder-techniques RPM packge on Linux

amousset:
https://www.rudder-project.org/redmine/issues/7714


peckpeck:
OK, merging

PR #837, opened 2016-01-08
Fixes #7739: Invalid reporting on windows for clockconfiguration if Hardware Clock is not set

ncharles:
https://www.rudder-project.org/redmine/issues/7739


jooooooon:
:+1:

PR #838, opened 2016-01-11
Fixes #7635: syslog-ng's loghost regex does not match with set line

peckpeck:
https://www.rudder-project.org/redmine/issues/7635


jooooooon:
This approach is incomplete - it doesn't remove old lines, and will be hard to adapt to upper versions. I've proposed another approach in #839.

PR #839, opened 2016-01-11
Fixes #7635: syslog-ng's loghost regex does not match with set line

jooooooon:
https://www.rudder-project.org/redmine/issues/7635


techniques/system/common/1.0/promises.st
```diff
@@ -753,9 +757,14 @@ bundle edit_line edit_syslog_conf_file(line_to_add, pattern_to_remove)
 
   delete_lines:
       "${pattern_to_remove}";
+
+      # Remove any unexpanded variables that may have slipped through due to old bugs or race conditions at installation time
       "\$\(syslog_ng_conf\)";
       "\$\{syslog_ng_conf\}";
 
+      # Remove the syslog config when the loghost was just called "loghost" (now called "rudder_loghost")
+      "filter\ f\_local\_rudder\{facility\(local6\)\ and\ program\(\"rudder\.\*\"\)\;\}\;destination\ loghost\ \{(tcp|udp)\(\"[^\"]+\".*";
```
peckpeck: Why is there backslash in front of spaces and underscores ?

peckpeck:
Aside of that, everything is good

techniques/system/common/1.0/promises.st
jooooooon: Dunno. It's copy pasted from the regex we used above and we know works...

jooooooon:
PR updated

peckpeck:
OK, merging

PR #843, opened 2016-01-14
Fixes #7761: Error in environment-variables.cf on Centos 5

amousset:
https://www.rudder-project.org/redmine/issues/7761


jooooooon:
I'm pretty sure we can work around this with a bit of sed/grep/awk...

jooooooon:
I've proposed a fix in #844

PR #845, opened 2016-01-15
Fixes #7774: Invalid promises genration after #7761, due to string template interpretaion

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7774


peckpeck:
OK, merging

PR #836, opened 2015-12-31
Fixes #7696: Missing reporting changes after #7644

amousset:
https://www.rudder-project.org/redmine/issues/7696

PR #848, opened 2016-01-15
Fixes #7788: Permissions of rudder tools are not preserved on policy servers - Missing change in initial promises

amousset:
https://www.rudder-project.org/redmine/issues/7788


amousset:
Apply #829 and #836 to initial promises.

PR #849, opened 2016-01-15
Fixes #7785: Rudder inventory don't produce error report

amousset:
https://www.rudder-project.org/redmine/issues/7785

### DIFF
--- a/initial-promises/node-server/promises.cf
+++ b/initial-promises/node-server/promises.cf
@@ -492,7 +492,7 @@ body agent control
     #define here some environment variables
         environment => { "DEBIAN_FRONTEND=noninteractive" };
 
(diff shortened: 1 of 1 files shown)


amousset:
`could_not_download_uuid` is not in the abortclasses in technique, but only in initial-promises. @ncharles do you know if there is a reason for this?

jooooooon:
I think it may be to prevent the agent from trying to send an inventory if it can't contact it's server

peckpeck:
Then adding such a tech before sending the inventory is a better solution.

ncharles:
Sorry, I didn't see the question. The goal of having this class was indeed to prevent bogus inventory to be sent. But indeed, now promises changed, and this should not plainly abort. Merging

PR #851, opened 2016-01-19
Fixes #3741: Technique \"Name resolution\": Regexp to check value of Nameserver does not allow the use of variable

jooooooon:
https://www.rudder-project.org/redmine/issues/3741


jooooooon:
Some notes:

  # This will need to be applied to technique version 3.0 also, which is in 3.0+ branches. Logged in http://www.rudder-project.org/redmine/issues/7806.
  # This only allows to use a variable as the *entire* contents of the field. I think that's not very annoying for the IP addresses, but could be quite annoying for the domain name field. For existing versions, I don't want to change this too much, though, at the risk of changing expected behaviour in a minor upgrade.

PR #852, opened 2016-01-19
Fixes #7796: Rudder agent does not detect when promise copy failed

amousset:
https://www.rudder-project.org/redmine/issues/7796


ncharles:
Ok ! this is interesting that the timestamp is not in initial promises though

PR #856, opened 2016-01-21
Fixes #7813: Reporting in server-components missing when restarting some services (ie jetty)

jooooooon:
https://www.rudder-project.org/redmine/issues/7813


jooooooon:
PR updated

jooooooon:
PR updated

PR #857, opened 2016-01-21
Fixes #7821: Unexpected reports on \"Check jetty process\" component of system techniques

jooooooon:
https://www.rudder-project.org/redmine/issues/7821


peckpeck:
OK, merging

PR #859, opened 2016-01-26
Fixes #7741: date of report may be invalid on windows

ncharles:
https://www.rudder-project.org/redmine/issues/7741


ncharles:
getDate gave invalid timezone, so i had to replace it by powershell
I cannot simply run powershell script  (I can use the powershell shell instead of useshell or noshell, but it doesn't work on non-nova system), so I ran a one-liner, and remove getDate.bat altogether

peckpeck:
OK merging

jooooooon:
:+1:

PR #860, opened 2016-01-31
Fixes #7489: Add all timezones and custom input

matya:
Changed GUI to have a dropdown of all possible timezones and also
the option to include a "Custom" Input box, useful for generic
variable defined TZ's.


techniques/systemSettings/misc/clockConfiguration/3.0/clockConfiguration.st
```diff
@@ -166,26 +88,26 @@ bundle agent check_clock_configuration
         classes => kept_if_else("ntpconf_kept", "repaired_ntpconf", "not_repaired_ntpconf");
 
     # Copy the correct localtime file (distro independant)
-    !clock_timezone_noedit.!windows.clock_vardef::
+    linux.!clock_timezone_noedit::
```
matya: This was made to be freebsd-compatible

techniques/systemSettings/misc/clockConfiguration/3.0/tz_vars.st
```diff
@@ -0,0 +1,932 @@
...
+# 
+#  You should have received a copy of the GNU General Public License
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# 
+
+bundle agent tz_variables
+{
+  vars:
+
+    ## The values for some deprecated fields we plan to phase out in future versions
```
matya: These fields below are required to have the technique's behavior stay the same on an unchanged input from the drop-down menu, still using the Countries instead of the new City-List.
Could be removed in technique version 4.0?

techniques/systemSettings/misc/clockConfiguration/3.0/clockConfiguration.st
```diff
@@ -30,6 +30,11 @@ bundle agent check_clock_configuration
   vars:
       "ntpServers" slist => {&CLOCK_NTPSERVERS: { "&it&"};separator=", "&};
 
+      "target_timezone" string => ifelse( strcmp("&CLOCK_TIMEZONE&", "custom"), "&CLOCK_TIMEZONE_CUSTOM&", "&CLOCK_TIMEZONE&" );
```
matya: here we find out if we have chosen to use the "Custom" input field or we should go with the actual value of the Drop-Down menu.

techniques/systemSettings/misc/clockConfiguration/3.0/clockConfiguration.st
matya: as I noticed, this is also the way to do it on freebsd.

techniques/systemSettings/misc/clockConfiguration/3.0/changelog
```diff
@@ -4,3 +4,7 @@
  -- Benoît Peccatte <benoit.peccatte@normation.com> Fri Oct 17 14:10:43 CEST 2014
   * Version 3.0
   ** Use rudder_common_report instead of reports:
+ -- Janos Mattyasovszky <mail@matya.hu> Sun Jan 31 01:04:00 CET 2016
+  * Version 3.0
```
ncharles: I think there's been an error there. Adding new stuff to a technique should make it a new version (3.1)

matya: Did not find any detailed information about when to increment the minor version.. . 
Basically no new stuff was added in the means of features, only the representation in the gui was changed. Should I make it 3.1 then?

matya: Did not find any detailed information about when to increment the minor version.. . 
Basically no new stuff was added in the means of features, only the representation in the gui was changed. Should I make it 3.1 then?

ncharles: you are adding tons of timezones - this does sound like a new feature :)

matya:
@ncharles changed to 3.1, including a copy from 3.0, so the diff still can be reviewed.

matya:
ping @ncharles @peckpeck ... could anybody give some feedback on this?

peckpeck:
Thank you @matya 
Merging

PR #866, opened 2016-02-15
Fixes #7924: Several techniques are missing explicit default values for booleans

jooooooon:
https://www.rudder-project.org/redmine/issues/7924


peckpeck:
OK, merging

PR #872, opened 2016-02-22
Fixes #7953: Technique \"Download a file from the shared folder\" is missing a default value for permissions in version >= 1.6

jooooooon:
https://www.rudder-project.org/redmine/issues/7953


peckpeck:
OK, merging

PR #873, opened 2016-02-22
Fixes #7955: Technique \"Download a file from the shared folder\" is missing a default value for permissions in version 2.0

jooooooon:
https://www.rudder-project.org/redmine/issues/7955


peckpeck:
OK, merging

PR #876, opened 2016-02-22
Fixes #7962: \"Download a file from shared folder\" technique doesn't have an explicit default value for \"Compare method\" (was mtime, now digest)

jooooooon:
https://www.rudder-project.org/redmine/issues/7962


peckpeck:
OK, merging

PR #881, opened 2016-02-26
Fixes #7983: Add an option for MOTD Technique to display MOTD before login

ncharles:
https://www.rudder-project.org/redmine/issues/7983


jooooooon:
Since changing /etc/issue is not really changing the MOTD, I propose changing the name of the technique in this new version to "MOTD and pre-login banner configuration"

ncharles:
PR updated

PR #884, opened 2016-02-29
Fixes #7994:  \"Download a file from shared folder\" technique doesn't have an explicit default value for \"Compare method\" (was mtime, now digest) - 3.0 branch

jooooooon:
https://www.rudder-project.org/redmine/issues/7994


jooooooon:
PR updated

peckpeck:
OK, merging

PR #886, opened 2016-03-01
Fixes #8009: Fix statement about NTP Servers in clockSettings 3.1

matya:


ncharles:
Good catch, thank you !

PR #888, opened 2016-03-01
Fixes #8011: Technique check repository gpg key doesn't follow the normal ordering

ncharles:
https://www.rudder-project.org/redmine/issues/8011


peckpeck:
OK, merging

PR #892, opened 2016-03-03
Fixes #8034: When we manage twice a same package in rpm package management, we don't get the second report

ncharles:
https://www.rudder-project.org/redmine/issues/8034


ncharles:
WIP
must do also previous versions

ncharles:
ha, actually previous version don't use the methods: for reporting, so good to merge

ncharles:
PR updated

ncharles:
PR updated

amousset:
Ok, merging.

PR #895, opened 2016-03-11
Fixes #8039: Broken inventory on root server after install

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8039


amousset:
OK, merging

PR #896, opened 2016-03-14
Fixes #8068: Technique jobScheduler v2.0 doesn't schedule

peckpeck:
https://www.rudder-project.org/redmine/issues/8068


amousset:
Ok, merging.

PR #898, opened 2016-03-17
Fixes #8079: Missing reporting in UserManagement when creating a user

amousset:
https://www.rudder-project.org/redmine/issues/8079

### DIFF
--- a/techniques/systemSettings/userManagement/userManagement/5.0/userManagement.st
+++ b/techniques/systemSettings/userManagement/userManagement/5.0/userManagement.st
@@ -249,7 +249,7 @@ bundle agent check_usergroup_user_parameters
 
       ## Added (Repaired)
       "@@userGroupManagement@@result_repaired@@${usergroup_directive_id[${usergroup_user_index}]}@@Users@@${usergroup_user_login[${usergroup_user_index}]}@@${g.execRun}##${g.uuid}@#The user ${usergroup_user_login[${usergroup_user_index}]} ( ${repname[${usergroup_user_index}]} ) has been added to the system"
-        ifvarclass => "!usermanagement_user_exists_${usergroup_user_index}.usermanagement_user_update_${usergroup_user_index}.usermanagement_login_add_${usergroup_user_index}_repaired";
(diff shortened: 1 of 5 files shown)


techniques/systemSettings/userManagement/userManagement/5.0/userManagement.st
```diff
@@ -249,7 +249,7 @@ bundle agent check_usergroup_user_parameters
 
       ## Added (Repaired)
       "@@userGroupManagement@@result_repaired@@${usergroup_directive_id[${usergroup_user_index}]}@@Users@@${usergroup_user_login[${usergroup_user_index}]}@@${g.execRun}##${g.uuid}@#The user ${usergroup_user_login[${usergroup_user_index}]} ( ${repname[${usergroup_user_index}]} ) has been added to the system"
-        ifvarclass => "!usermanagement_user_exists_${usergroup_user_index}.usermanagement_user_update_${usergroup_user_index}.usermanagement_login_add_${usergroup_user_index}_repaired";
+        ifvarclass => "usermanagement_user_exists_${usergroup_user_index}.usermanagement_user_update_${usergroup_user_index}.usermanagement_login_add_${usergroup_user_index}_repaired";
```
ncharles: Hum.
the whole idea of the !user_exists is to know if the user existed before or not (and it is repaired if user didn't exists before)
I guess the issue is that user didn't exist, user was create, user exist, do no reporting

I think a better fix would be to have a
"user_didn_t_exist" class which is not => userexists(...)

amousset: We already have the "usermanagement_login_add_${usergroup_user_index}_repaired" class which can only be set if the useradd command was executed, which only happens when the user does not exist in pass2.
Do you think it is enough or we should add another dedicated class?

amousset: ping @ncharles

ncharles: thank you for the ping
You are correct, repaired is only set if user is created

PR #900, opened 2016-03-29
Fixes #8080: Missing reporting in UserManagement 6.0 when creating a user

amousset:
https://www.rudder-project.org/redmine/issues/8080


amousset:
@ncharles Same as #898 bu for the 6.0 version.

PR #901, opened 2016-03-31
Fixes #8134: Do not use proxy when calling localhost in techniques

peckpeck:
https://www.rudder-project.org/redmine/issues/8134

### DIFF
--- a/initial-promises/node-server/server-roles/1.0/alive-check.cf
+++ b/initial-promises/node-server/server-roles/1.0/alive-check.cf
@@ -54,7 +54,7 @@ bundle agent generic_alive_check(app)
 
     root_server::
 
-      "/usr/bin/curl -s ${site_to_check} |/bin/grep -q OK"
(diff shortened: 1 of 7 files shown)


initial-promises/node-server/server-roles/1.0/alive-check.cf
```diff
@@ -54,7 +54,7 @@ bundle agent generic_alive_check(app)
 
     root_server::
 
-      "/usr/bin/curl -s ${site_to_check} |/bin/grep -q OK"
+      "/usr/bin/curl --proxy -s ${site_to_check} |/bin/grep -q OK"
```
ncharles: should be --proxy ''
isn't it ?

matya: it should.

peckpeck:
PR updated

amousset:
Ok, merging.

PR #904, opened 2016-04-04
Fixes #8149: On Windows, a file c:\Program is created

ncharles:
https://www.rudder-project.org/redmine/issues/8149

### DIFF
--- a/techniques/system/inventory/1.0/fusionAgent.st
+++ b/techniques/system/inventory/1.0/fusionAgent.st
@@ -285,7 +285,7 @@ bundle agent listInstalledVM
 &if(NOVA)&
     windows.VirtualBoxInstalled::
       "\"${virtualMachines.virtual_box_install_path}VBoxManage.exe\""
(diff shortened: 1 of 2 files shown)


techniques/system/inventory/1.0/fusionAgent.st
```diff
@@ -285,7 +285,7 @@ bundle agent listInstalledVM
 &if(NOVA)&
     windows.VirtualBoxInstalled::
       "\"${virtualMachines.virtual_box_install_path}VBoxManage.exe\""
-        args => "-q list vms > ${virtualMachines.VBoxListFile}",
+        args => "-q list vms > \"${virtualMachines.VBoxListFile}\"",
```
amousset: Isn't the escaping also needed at lines 280 and 305?

ncharles: no, because 280 and 305 are not windows, hence no Program Files

amousset:
Ok, merging.

PR #906, opened 2016-04-07
Fixes #8159: Fix copy_from bodies for Rudder files

amousset:
https://www.rudder-project.org/redmine/issues/8159


amousset:
PR updated

ncharles:
I love that.

ncharles:
This look great. If you confirm that you tested it (with relays) and that it does work as expected, i'll merge it :)

PR #910, opened 2016-04-21
Fixes #8108: The start at boot does not work in the openssh technique on systemd OSes

amousset:
https://www.rudder-project.org/redmine/issues/8108

### DIFF
--- a/techniques/systemSettings/remoteAccess/sshConfiguration/5.0/main.st
+++ b/techniques/systemSettings/remoteAccess/sshConfiguration/5.0/main.st
@@ -40,14 +40,6 @@ bundle agent rudder_openssh_server_configuration(class_prefix, service_name, par
      "rudder_openssh_server_service_name"
         string => "sshd";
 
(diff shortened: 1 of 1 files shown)


jooooooon:
This looks good to me. However, this bug is also present in version 4.0 of the technique (but not 3.0), so please also fix it there in this same PR.

amousset:
The 2.11/4.0 does not use ncf, and handling systemd requires to manage it manually in the technique.

However, the Rudder 3.0/Technique 4.0 uses ncf. Should I:
* Backport the ncf version to 2.11
* Fix in the 4.0 technique starting from Rudder 3.0 and leave the 2.11 version unchanged (given that we do not support any systemd systems in 2.11).

jooooooon:
Since the version 4.0 in branch 2.11 is not broken, let's not mess with it. Just fix the version 3.0 from branch 3.0.

amousset:
Opened #916 for the 3.0 branch.

PR #911, opened 2016-04-22
Fixes #8198: permissions technique tries to set a group even when we do not ask for it

amousset:
https://www.rudder-project.org/redmine/issues/8198


jooooooon:
Looks good, but also needs fixing in versions 2.0 and 2.1 in the 3.0 branch. I created https://www.rudder-project.org/redmine/issues/8203 to keep track of this.

PR #914, opened 2016-04-25
Fixes #8121: Clear package caches when repairing a package repository

amousset:
https://www.rudder-project.org/redmine/issues/8121

### DIFF
--- a/techniques/applications/zypperPackageManagerRepositories/1.0/zypper-repositories-management.st
+++ b/techniques/applications/zypperPackageManagerRepositories/1.0/zypper-repositories-management.st
@@ -75,6 +75,12 @@ bundle agent zypper_repositories_management
     SuSE::
       "repos_${zypper_index}" usebundle => rudder_common_reports_generic("Zypper Repositories Management", "zypper_repo_${zypper_index}", "${zypper_uuid[${zypper_index}]}", "Repositories", "${zypper_name[${zypper_index}]}", "Repository ${zypper_name[${zypper_index}]}");
 
+      # Clear package caches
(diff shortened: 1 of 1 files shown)


peckpeck:
This looks like a bug and not a user story, moreover the issue is missing a description of a reason of this change, what problem does it fix ?

amousset:
I updated the issue description.

techniques/applications/zypperPackageManagerRepositories/1.0/zypper-repositories-management.st
```diff
@@ -75,6 +75,14 @@ bundle agent zypper_repositories_management
     SuSE::
       "repos_${zypper_index}" usebundle => rudder_common_reports_generic("Zypper Repositories Management", "zypper_repo_${zypper_index}", "${zypper_uuid[${zypper_index}]}", "Repositories", "${zypper_name[${zypper_index}]}", "Repository ${zypper_name[${zypper_index}]}");
 
+      # Clear package caches
+      "remove_caches" usebundle => file_remove("/var/rudder/cfengine-community/state/software_packages.csv"),
```
peckpeck: Doesn't this contain the installed package list ?
If so, why would you need to update it after a repository change ?

amousset: These files are created by the agent when they do not exist.

amousset: We do not need to remove, this file, indeed.

amousset:
PR updated

peckpeck:
OK, merging

PR #915, opened 2016-04-25
Fixes #8208: check_cron_daemon conflicts with cronManagement

matya:
Make sure cronManagement's reports are only triggered when its local
class has determined the need to restart the service, and not a fake
class from check_cron_daemon


peckpeck:
Hello @matya since @ncharles is very busy now, I will review this one

peckpeck:
Good, merging
Thanks @matya

PR #917, opened 2016-05-06
Fixes #8243: Clear package caches when repairing a package repository for yum and apt

amousset:
https://www.rudder-project.org/redmine/issues/8243


peckpeck:
Where is the yum case ?

peckpeck:
Be aware that there are new versions of the technique in newer rudder versions

peckpeck:
Good, merging

PR #919, opened 2016-05-09
Fixes #8100: Double reporting for group creation

jooooooon:
https://www.rudder-project.org/redmine/issues/8100


peckpeck:
OK, merging

PR #921, opened 2016-05-10
Fixes #8263: Load environment variables from /etc/profile 

amousset:
https://www.rudder-project.org/redmine/issues/8263


jooooooon:
:+1:

PR #924, opened 2016-05-17
Fixes #8161: Crontab promise with the Windows agent

ncharles:
https://www.rudder-project.org/redmine/issues/8161


initial-promises/node-server/common/1.0/process_matching.cf
```diff
@@ -55,7 +55,7 @@ bundle agent process_matching
       # (see http://www.rudder-project.org/redmine/issues/3925 and http://www.rudder-project.org/redmine/issues/3930).
       "/etc/cron.d/rudder-agent-uuid" delete => tidy;
 
-    community_edition.!aix::
+    community_edition.!aix.!windows::
```
peckpeck: Doesn't "community_edition" imply "!windows" ?

ncharles: implicitely, but i'd rather be safe and cover the unexpected case when a community version of an agent would exist on windows (like, cygwin)

peckpeck:
OK, merging

PR #928, opened 2016-05-17
Fixes #8313: Rename poorly named tool \"signature.sh\"  and use it in rudder-techniques

jooooooon:
https://www.rudder-project.org/redmine/issues/8313

### DIFF
--- a/techniques/system/inventory/1.0/fusionAgent.st
+++ b/techniques/system/inventory/1.0/fusionAgent.st
@@ -564,7 +564,8 @@ bundle agent signInventories
   vars:
 
     !windows::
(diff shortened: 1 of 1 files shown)


techniques/system/inventory/1.0/fusionAgent.st
```diff
@@ -573,12 +574,11 @@ bundle agent signInventories
 
   files:
 
-    !windows::
+    !windows.sign_script_exists::
```
jooooooon: This is a simple optimisation - ifvarclass is slower than prefixed classes.

peckpeck:
OK, merging

PR #929, opened 2016-05-17
Fixes #8316: Add a flag file to tell the agent not to restart cf-serverd

jooooooon:
https://www.rudder-project.org/redmine/issues/8316

### DIFF
--- a/techniques/system/common/1.0/update.st
+++ b/techniques/system/common/1.0/update.st
@@ -108,6 +108,9 @@ bundle common server_info
       "cfserved" string => ifelse(fileexists("${policy_server_file}"),
                                   readfile("${policy_server_file}", 2048),
                                   "rudder");
+
+  classes:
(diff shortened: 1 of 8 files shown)


jooooooon:
Commit modified

peckpeck:
I'm OK with this patch, but shouldn't it be on older branches too ?
Since the technique can on a recent agent from an older server.

peckpeck:
This PR is nor mergeable anymore

techniques/system/common/1.0/update.st
```diff
@@ -367,7 +367,7 @@ bundle agent update_action
         action  => u_ifwin_bg,
         classes => outcome("executor");
 
-    start_server::
+    !should_disable_server.start_server::
```
ncharles: i think it should go as far as stopping effectively the cf-serverd component

jooooooon:
Commit modified

PR #931, opened 2016-05-18
Fixes #8045: Using Rudder server 3.x + rudder agent 2.11.x, a node doesn't properly detect its change of relay server

ncharles:
https://www.rudder-project.org/redmine/issues/8045


peckpeck:
OK, merging

PR #933, opened 2016-05-18
Fixes #7328: Extra carriage return in policy_server.dat inventory failure

jooooooon:
https://www.rudder-project.org/redmine/issues/7328


jooooooon:
A better solution to this would be to use regextract, which would actually "trim" whitespace from the beginning of the value too. However, any solutions that require a function that returns a class (rather than a var) don't work in CFEngine 3.5 because of some problem with multiple passes.

I therefore resorted to using readstringarrayidx which works in all versions.

ncharles:
ok, merging

PR #934, opened 2016-05-19
Fixes #8336: Error on windows: agent tries to check if /usr/bin/env supports the -0 option

ncharles:
https://www.rudder-project.org/redmine/issues/8336


amousset:
Ok, merging.

PR #937, opened 2016-05-23
Fixes #6421: Messages can be dropped on the node, resulting in Unknown reports on the Web Interface

amousset:
https://www.rudder-project.org/redmine/issues/6421


techniques/system/common/1.0/promises.st
```diff
@@ -494,7 +494,7 @@ bundle agent check_log_system
 
     !windows.rsyslogd.!policy_server::
       "/etc/rsyslog.d/rudder-agent.conf"
-        edit_line => append_if_no_lines("#Rudder log system${const.n}if $syslogfacility-text == 'local6' and $programname startswith 'rudder' then @@${server_info.cfserved}:&SYSLOGPORT&${const.n}if $syslogfacility-text == 'local6' and $programname startswith 'rudder' then ~"),
+        edit_line => append_if_no_lines("#Rudder log system${const.n}$RepeatedMsgReduction off${const.n}if $syslogfacility-text == 'local6' and $programname startswith 'rudder' then @@${server_info.cfserved}:&SYSLOGPORT&${const.n}if $syslogfacility-text == 'local6' and $programname startswith 'rudder' then ~"),
```
jooooooon: This only addresses the rsyslog case. Is there no equivalent for syslog-ng? syslog?

peckpeck: What about existing configuration ?
This will append a new line.

jooooooon: I thought that, but actually it won't, because of the line a bit below:

```
edit_defaults => empty_backup,
```

jooooooon:
Ping @amousset

amousset:
I'm doing some tests on syslog-ng with the `suppress` option, but it seems disabled by default (at least on SLES). For syslogd it seems quite hazardous (http://www.syslog.org/forum/syslog-and-syslogd/disable-%27last-message-repeated%27-compression-on-lin/).

amousset:
The suppress option is not supported by the syslog-ng version in SELS11 repos. I am trying to ensure we won't have issue with the rsyslog option on old versions as well, I could not find the precise version where it was included for now.

ncharles:
I tried to locate when it was first added also, without success. I seem to recall that it was there at least in 3.8 (but not sure at all)

amousset:
Rsyslog was apparently introduced in RHEL repos in RHEL5.2 (http://lists.adiscon.net/pipermail/rsyslog/2008-June/000836.html), and was a 2.0 version, which seems not to support this option (http://rpm.pbone.net/index.php3/stat/45/idpl/11534950/numer/5/nazwa/rsyslog.conf)

amousset:
We can:
* Find when the RepeatedMsgReduction was introduced
* Adapt check-rsyslog-version to be able to check for this version
* Add RepeatedMsgReduction only when compatible

The is no simple way to disable message deduplication on syslogd on Linux (on BSD there is a command-line option). For syslog-ng, it was introduced in a recent version (for now, unknown too...), so we would need the same mechanism. By the way, it seems at the time it was introduced, all Linux distributions had already switched to rsyslog by default.

jooooooon:
Great news, @amousset ! Let's focus on rsyslog, then, since that seems to be the most widely used, and the only one we've actually experienced the problem with.

amousset:
Commit modified

jooooooon:
This looks great!

techniques/system/common/1.0/promises.st
```diff
@@ -549,6 +559,14 @@ bundle agent check_log_system
 
     aix.syslogd_repaired::
       "/usr/bin/refresh -s syslogd";
+      
+    rsyslogd.check_rsyslog_version_present::
+      # 5.6.4 is the first version mentionning RepeatedMsgReduction in http://www.rsyslog.com/change-log/
+      # The option is probably older though.
+      "${g.rudder_tools}/check-rsyslog-version 5.6.4"
```
ncharles: hum, we have a conflict there. Classes defined by module are global, so if we hve a version 5.6.5, we'll have both rsyslog_version_ok and rsyslog_version_not_ok defined

PR #939, opened 2016-05-24
Fixes #6487: Avoid waiting on lock when calling yum

ncharles:
https://www.rudder-project.org/redmine/issues/6487


peckpeck:
OK, merging

PR #946, opened 2016-05-25
Fixes #8384: Adapt check-techniques script to new name for rudder-stdlib-core

jooooooon:
https://www.rudder-project.org/redmine/issues/8384


peckpeck:
Good, merging

PR #949, opened 2016-05-25
Fixes #8387: check-techniques script doesn't properly exit non-zero on errors

jooooooon:
https://www.rudder-project.org/redmine/issues/8387


peckpeck:
OK, merging

PR #950, opened 2016-05-26
Fixes #8389: Merge rudder_lib from initial promises and system technique - adapt failsafe.cf for new rudder-stdlib-core.cf name

jooooooon:
https://www.rudder-project.org/redmine/issues/8389


peckpeck:
Good, merging

PR #951, opened 2016-05-26
Fixes #8383: Merge remaining files from initial promises and system technique

peckpeck:
https://www.rudder-project.org/redmine/issues/8383


initial-promises/node-server/promises.cf
```diff
@@ -375,63 +375,6 @@ body process_select cron_bin
   process_result => "command";
 }
 
-#######################################################
-# Check if the cron daemon is running
-# This only works with unix flavoured systems too
-
-bundle agent check_cron_daemon
```
jooooooon: This bundle has moved to ./common/1.0/cron_setup.st, but that file is not yet included in initial-promises. Can you add it so we know we have an equivalent add to this remove?

peckpeck:
Commit modified

peckpeck:
Commit modified

PR #957, opened 2016-05-27
Fixes #8390: Merge remaining files from initial promises and system technique - bis

peckpeck:
https://www.rudder-project.org/redmine/issues/8390

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,12 +21,14 @@ WGET := $(if $(PROXY), http_proxy=$(PROXY) ftp_proxy=$(PROXY)) /usr/bin/wget -q
 all: rudder-templates-cli.jar
 	cp techniques/system/common/1.0/rudder-stdlib.cf initial-promises/node-server/common/1.0/
 	sed -i -e 's/.*TRACKINGKEY.*/  "TRACKINGKEY": "hasPolicyServer-root@@common-root@@00",/' variables.json
-	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0/ techniques/system/common/1.0/rudder-stdlib-core.st
-	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0/ techniques/system/common/1.0/cf-served.st
-	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0/ techniques/system/common/1.0/internal_security.st
-	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0/ techniques/system/common/1.0/rudder_lib.st
-	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0/ techniques/system/common/1.0/site.st
-	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0/ techniques/system/common/1.0/cron_setup.st
+	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0 techniques/system/common/1.0/rudder-stdlib-core.st
+	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0 techniques/system/common/1.0/cf-served.st
+	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0 techniques/system/common/1.0/internal_security.st
+	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0 techniques/system/common/1.0/rudder_lib.st
+	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0 techniques/system/common/1.0/site.st
+	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0 techniques/system/common/1.0/update.st
+	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server techniques/system/common/1.0/failsafe.st
+	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/common/1.0 techniques/system/common/1.0/cron_setup.st
 	java -jar rudder-templates-cli.jar --outext '' --outdir initial-promises/node-server/common/cron techniques/system/common/1.0/rudder_agent_community_cron.st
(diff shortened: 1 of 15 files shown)


Makefile
```diff
@@ -57,6 +59,7 @@ all: rudder-templates-cli.jar
 	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/server-roles/1.0 techniques/system/server-roles/1.0/servers-by-role.st
 	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/server-roles/1.0 techniques/system/server-roles/1.0/service-check.st
 	java -jar rudder-templates-cli.jar --outext .cf --outdir initial-promises/node-server/server-roles/1.0 techniques/system/server-roles/1.0/technique-reload.st
+	sed -i -e 's/.*TRACKINGKEY.*/  "TRACKINGKEY": "root-DP@@root-distributePolicy@@00",/' variables.json
```
jooooooon: Why is this necessary, with no more calls to rudder-templates-cli.jar after it?

initial-promises/node-server/common/1.0/update.cf
```diff
@@ -186,155 +237,198 @@ bundle agent update_action
...
+        comment           => "Update the common Rudder ncf instance";
+
+      "${g.rudder_ncf}/local"
+        copy_from         => remote("${server_info.cfserved}", "${g.rudder_ncf_origin_local}"),
+        depth_search      => recurse_ignore("inf", @{g.excludedreps}),
+        perms             => u_mog("644", "root", "0"),
+        action            => immediate,
+        move_obstructions => "true",
+        classes           => success("rudder_ncf_local_updated", "rudder_ncf_local_update_error", "rudder_ncf_local_updated_ok"),
+        comment           => "Update the local Rudder ncf instance";
```
jooooooon: This means we will copy/update ncf with initial-promises, therefore at times when the node this is running on has not been accepted. In all likelihood, this means there will be times when we can update ncf (since the cf-serverd ACLs allow access to ncf for all nodes in the allowed networks - except for local/50_techniques). If there is a significant change in ncf and the initial promises use ncf, this could break initial promises, and never get fixed.

I think this needs some thought.

initial-promises/node-server/common/1.0/update.cf
```diff
@@ -186,155 +237,198 @@ bundle agent update_action
...
     rudder_tools_update_error::
       "${g.rudder_tools_updated}"
         delete => tidy,
         comment => "Deleting ${g.rudder_tools_updated} as the tools couldn't be updated";
 
+  methods:
+    pass3.server_ok::
+      "any" usebundle => rudder_common_report("Common", "log_repaired", "hasPolicyServer-root@@common-root@@00", "Update", "None", "Started the server (cf-serverd)");
+    pass3.executor_ok::
+      "any" usebundle => rudder_common_report("Common", "log_repaired", "hasPolicyServer-root@@common-root@@00", "Update", "None", "Started the scheduler (cf-execd)");
```
jooooooon: These logs should be in the update_reports bundle below. They will break reporting by being here (since this is run by failsafe.cf and our "start run" message is sent out at the beginning of promises.cf)

initial-promises/node-server/failsafe.cf
```diff
@@ -103,147 +99,3 @@ bundle common g
...
-          file_result => "!leaf_name";
-}
-
-body file_select by_fullname(names)
-# @brief Select files that match `names`
-# @param names A regular expression
-{
-          path_name  => { @(names)};
-          file_result => "path_name";
-}
```
jooooooon: What happened here? It looks like this file has been truncated.

Makefile
peckpeck: It's just handy.
When you run the makefile, variables.json stays the same, so I have no warning about a changed file anymore.

initial-promises/node-server/failsafe.cf
peckpeck: Just a dangling &endif& in failsafe.st, will remove it.

initial-promises/node-server/common/1.0/update.cf
peckpeck: This is a bug that will be solved by http://www.rudder-project.org/redmine/issues/8422

initial-promises/node-server/common/1.0/update.cf
peckpeck: You're right, we should disable this in initial promises

peckpeck:
Commit modified

peckpeck:
PR rebased

initial-promises/node-server/failsafe.cf
```diff
@@ -104,53 +100,60 @@ bundle common g
...
 
+
+      "rudder_tools_updated_filename" string => "rudder_tools_updated";
+      "rudder_tools_updated_origin" string => "${rudder_tools_origin}/${rudder_tools_updated_filename}";
+      "rudder_tools_updated"        string => "${rudder_tools}/${rudder_tools_updated_filename}";
+
+      # The time at which the execution started
+    windows.!cygwin::
+      "execRun"                  string => execresult("c:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -Command \"'{0:yyyy-MM-dd HH:mm:sszzz}' -f (get-date)\"", "noshell");    android::
+      "execRun"                  string => execresult("/system/xbin/date \"+%Y-%m-%d %T+02:00\"", "noshell");
```
jooooooon: Something has gone wrong here: there should be a line with "android::" just above this one.

techniques/system/common/1.0/update.st
```diff
@@ -268,6 +267,7 @@ bundle agent update_action
         file_select       => exclude("${rudder_promises_timestamp_tmp}"),
         classes           => success("config", "no_update", "config_ok");
 
+&if(INITIAL)&
```
jooooooon: I think this should be "*not* INITIAL", otherwise we won't update ncf in the main promises anymore!

jooooooon:
Aside from those two comments this is looking great.

peckpeck:
PR rebased

peckpeck:
Commit modified

peckpeck:
PR rebased

initial-promises/node-server/failsafe.cf
jooooooon: Still missing.

peckpeck: Oh, there is a BIG warning we should have on string template, it always removes the line after an &endif& !!!

peckpeck:
Commit modified

jooooooon:
That's kinda scary... could we add a test in scripts/check-techniques.sh to make sure all "&endif&" lines are followed by a blank line?

PR #959, opened 2016-05-30
Fixes #8419: rpmPackageInstallation 7.0 uses the wrong version number for it's data file

jooooooon:
https://www.rudder-project.org/redmine/issues/8419


peckpeck:
OK, merging

PR #960, opened 2016-05-30
Fixes #8421: Invalidation of previous package state and of package status cache does not work on rpmPackageInstallation 5.0 5.1 6.0 6.1 7.0

jooooooon:
https://www.rudder-project.org/redmine/issues/8421


ncharles:
great catch !

PR #961, opened 2016-05-30
Fixes #8422: failsafe sometimes produce rudder reports

peckpeck:
https://www.rudder-project.org/redmine/issues/8422

### DIFF
--- a/techniques/system/common/1.0/update.st
+++ b/techniques/system/common/1.0/update.st
@@ -269,7 +269,7 @@ bundle agent update_action
       "CfengineNovaExec"
         service_policy => "start",
         service_method => u_bootstart,
-        classes        => outcome("executor"),
+        classes        => success("executor_ok", "executor_error", "executor_kept"),
         comment        => "Start the executor windows service now and at boot time";
 &endif&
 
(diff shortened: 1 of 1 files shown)


techniques/system/common/1.0/update.st
```diff
@@ -347,6 +340,10 @@ bundle agent update_reports
     (config|rudder_tools_updated|rudder_ncf_common_updated|rudder_ncf_local_updated|server_ok|executor_ok).!(rudder_promises_generated_error|rudder_tools_updated_error|rudder_tools_update_error|rudder_ncf_common_update_error|rudder_ncf_local_update_error|no_update)::
       "@@Common@@result_repaired@@&TRACKINGKEY&@@Update@@None@@${g.execRun}##${g.uuid}@#Rudder policy, tools or ncf instance were updated or CFEngine service restarted";
 
+    server_ok::
+      "@@Common@@log_repaired@@&TRACKINGKEY&@@Update@@None@@${g.execRun}##${g.uuid}@#Started the server (cf-serverd)";
+    executor_ok::
+      "@@Common@@log_repaired@@&TRACKINGKEY&@@Update@@None@@${g.execRun}##${g.uuid}@#Started the scheduler (cf-execd)";
```
jooooooon: These reports will never be printed because the server_ok and executor_ok classes in the previous bundle and not persistent, and this bundle is run in a separate process.

I think you'll need to also change the "outcome()" body classes to "success(if, else, kept)"

peckpeck:
Commit modified

techniques/system/common/1.0/update.st
```diff
@@ -269,32 +269,25 @@ bundle agent update_action
       "CfengineNovaExec"
         service_policy => "start",
         service_method => u_bootstart,
-        classes        => outcome("executor"),
+        classes        => success("executor_ok", "executor_kept", "executor_error"),
```
jooooooon: These classes are the wrong order. It should be repaired, error, kept (see above)

peckpeck:
Commit modified

PR #964, opened 2016-06-01
Fixes #8403: error: /default/check_generic_file_content/reports: Variable reference bracket mismatch

ncharles:
https://www.rudder-project.org/redmine/issues/8403


peckpeck:
Good, merging

PR #965, opened 2016-06-03
Fixes #8451: On Centos6.6, rsyslog is always restarted

amousset:
https://www.rudder-project.org/redmine/issues/8451

### DIFF
--- a/tools/check-rsyslog-version
+++ b/tools/check-rsyslog-version
@@ -1,9 +1,8 @@
 #!/bin/sh
 
-# The minimal rsyslog version we support is 5.7.1.
-# This utility checks if this condition is met and
-# returns the appropriate CFEngine class using the
-# module protocol
+# This script allows testing for a minimal rsyslog
+# version and define matching classes using
+# the module protocol.
 
 # rpmvercmp location
 RPMVERCMP="/var/rudder/cfengine-community/bin/rpmvercmp"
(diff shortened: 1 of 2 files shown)


amousset:
Still testing, do not merge for now.

ncharles:
can you tell when we can review ?

tools/check-rsyslog-version
```diff
@@ -15,14 +15,16 @@ else
   MINIMAL_RSYSLOG_VERSION="5.7.1"
 fi
 
+CANONIFIED_VERSION=$(echo ${MINIMAL_RSYSLOG_VERSION} | sed -r 's/\./_/g')
+
```
ncharles: comment at begining of module should be updated to tell it can take a parameter.
Plus, does this command work everywhere where we have rsyslog (aix?)

amousset:
Commit modified

amousset:
@ncharles Tested on 2.11/Debian 7

tools/check-rsyslog-version
amousset: Yes, the sed command is similar to the examples for AIX 5.3 (http://www.ibm.com/support/knowledgecenter/ssw_aix_53/com.ibm.aix.cmds/doc/aixcmds5/sed.htm), and the $() works in ksh (https://www.ibm.com/developerworks/aix/library/au-kornshellscripting/)

ncharles:
@amousset tested on centos6, works as expected.

PR #967, opened 2016-06-03
Fixes #8458: If using Rudder Server 3.1 and a node with agent 3.2, protocol used is TLS

ncharles:
https://www.rudder-project.org/redmine/issues/8458


amousset:
Tested on 2.11, everything is ok. Merging.

PR #981, opened 2016-07-06
Fixes #8585: Reporting for usermanagement in case of Policy to apply to this account "check only" is missing the Password component report

ncharles:
https://www.rudder-project.org/redmine/issues/8585


amousset:
Ok, merging.

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #989, opened 2016-07-13
Fixes #8599: UserManagement 6.0 on at least Debian fails to add user if the user's default group already exists

ncharles:
https://www.rudder-project.org/redmine/issues/8599


amousset:
Reproduced on CentOS 7.1, so the _debian_ condition is probably not the right one.

ncharles:
Commit modified

ncharles:
it's not mandatory on sles, but it doesn't break to use it

ncharles:
Commit modified

ncharles:
Commit modified

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #993, opened 2016-07-21
Fixes #8735: Allow overriding disable-agent detection

peckpeck:
https://www.rudder-project.org/redmine/issues/8735

### DIFF
--- a/initial-promises/node-server/promises.cf
+++ b/initial-promises/node-server/promises.cf
@@ -139,7 +139,7 @@ bundle agent check_disable_agent
 
       # Only define this class when we're ready to die - this is a special class name in "abortclasses"
       "should_not_continue"
-        expression => "should_disable_agent",
+        expression => "should_disable_agent.!ignore_disable_agent",
         ifvarclass => "abort_report_done",
         action     => immediate;
 
@@ -252,6 +252,12 @@ bundle agent check_red_button_status()
 # Check that CFengine services are up
 ###################################################
 bundle agent check_cf_processes
+{
+  methods:
+    !ignore_disable_agent::
+      "any" usebundle => check_cf_processes_running;
+}
+bundle agent check_cf_processes_running
 {
 
   processes:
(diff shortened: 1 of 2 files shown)


initial-promises/node-server/promises.cf
```diff
@@ -139,7 +139,7 @@ bundle agent check_disable_agent
 
       # Only define this class when we're ready to die - this is a special class name in "abortclasses"
       "should_not_continue"
-        expression => "should_disable_agent",
+        expression => "should_disable_agent.!run_disabled",
```
amousset: I don't really get the `run_disabled` name, I'd prefer something like `ignore_disable_agent`. We should also add a comment to explain the reason for this.

peckpeck:
Commit modified

techniques/system/common/1.0/promises.st
```diff
@@ -190,7 +190,7 @@ bundle agent check_disable_agent
 
       # Only define this class when we're ready to die - this is a special class name in "abortclasses"
       "should_not_continue"
-        expression => "should_disable_agent",
+        expression => "should_disable_agent.!ignore_disable_agent",
```
jooooooon: Hmm, just thinking about naming here. The command that uses this will be "rudder agent run --force", and we already have a "force_inventory" class to, well, force an inventory.

Can we name this force_agent_run instead? Or something similar, that starts with force_?

peckpeck: We don't have a rudder agent inventory --force and an associated class, maybe we should have a ignore_disable_inventory class instead

jooooooon: Well, "rudder agent inventory" already uses the force_inventory class. I think that the name is explicit enough that it should always run an inventory. I see no plausible case where you'd run "cf-agent -D force_inventory" and expect to be informed inventories are disabled.

jooooooon: OK, just saw your comment in #996, that makes for a reasonable case. I hadn't understood "disable inventory" is supposed to *prevent* all and any inventories, I imagined it was just to stop automatic (nightly) inventories. Let me think this over.

jooooooon: OK, after some thought, this *does* make sense, once you consider that "force_inventory" is a bit of a misnomer, but we're not going to change that now.

So, let's stick with ignore_disable_inventory and ignore_disable_agent class names.

peckpeck:
@jooooooon ok, so this PR doesn't need any change

PR #996, opened 2016-07-22
Fixes #8741: Disable inventories using a file

peckpeck:
https://www.rudder-project.org/redmine/issues/8741

### DIFF
--- a/initial-promises/node-server/inventory/1.0/fusionAgent.cf
+++ b/initial-promises/node-server/inventory/1.0/fusionAgent.cf
@@ -27,6 +27,17 @@
 
 bundle agent doInventory
 {
+  classes:
+    "disable_inventory" expression => fileexists("${g.rudder_disable_inventory_file}");
+  methods:
+    ignore_disable_inventory|!disable_inventory::
+      "any" usebundle => doInventory_always;
+  reports:
+    !(ignore_disable_inventory|!disable_inventory)::
+      "INFO: The file ${g.rudder_disable_inventory_file} is present. Rudder will not send an inventory. Use rudder agent inventory --force to override this temporarily.";
+}
+bundle agent doInventory_always
+{
   vars:
 
     uuid_succesfully_downloaded::
--- a/techniques/system/inventory/1.0/fusionAgent.st
+++ b/techniques/system/inventory/1.0/fusionAgent.st
@@ -55,6 +55,17 @@ bundle agent computeInventoryTime
 
 bundle agent doInventory
 {
+  classes:
+    "disable_inventory" expression => fileexists("${g.rudder_disable_inventory_file}");
+  methods:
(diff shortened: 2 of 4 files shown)


initial-promises/node-server/inventory/1.0/fusionAgent.cf
```diff
@@ -27,6 +27,14 @@
 
 bundle agent doInventory
 {
+  classes:
+    "disable_inventory" expression => fileexists("${g.rudder_disable_inventory_file}");
```
ncharles: shouldn't we have a verbose message here, saying the inventory is disabled ?

jooooooon: +1

techniques/system/inventory/1.0/fusionAgent.st
```diff
@@ -55,6 +55,14 @@ bundle agent computeInventoryTime
 
 bundle agent doInventory
 {
+  classes:
+    "disable_inventory" expression => fileexists("${g.rudder_disable_inventory_file}");
+  methods:
+    !disable_inventory::
+      "any" usebundle => doInventory_always;
+}
+bundle agent doInventory_always
```
jooooooon: I don't think you should separate this check into a different bundle, because there is a purposely-designed class "force_inventory" that is checked below. Just like #993 allows to bypass the disable-agent file, it would be good if this force_inventory class also bypasses the disable-inventory file.

Therefore, it would make more sense to integrate this check into the logic below.

peckpeck: The force_inventory is used twice, once in the"rudder agent inventory" command and in the packaging to run a normal inventory after an install or an upgrade.

In those case, the inventory should not be run if the disable-inventory file is present.

If we want to make this coherent with disable-agent, we should have a new parameter "rudder agent inventory --force" and split the force_inventory class into run_inventory_now and (really_)force_inventory

peckpeck:
Commit modified

peckpeck:
Commit modified

initial-promises/node-server/inventory/1.0/fusionAgent.cf
```diff
@@ -27,6 +27,17 @@
...
+  methods:
+    ignore_disable_inventory|!disable_inventory::
+      "any" usebundle => doInventory_always;
+  reports:
+    !(ignore_disable_inventory|!disable_inventory)::
+      "INFO: The file ${g.rudder_disable_inventory_file} is present. Rudder will not send an inventory.";
```
jooooooon: Can you add "Run 'rudder agent inventory --force' to override this temporarily." ?

peckpeck:
Commit modified

peckpeck:
OK, merging this PR

PR #1002, opened 2016-08-02
Fixes #8417: Merge promises.st from initial promises and system technique

peckpeck:
https://www.rudder-project.org/redmine/issues/8417

### DIFF
--- a/initial-promises/node-server/promises.cf
+++ b/initial-promises/node-server/promises.cf
@@ -23,105 +23,160 @@
 #######################################################
 bundle common rudder_roles
 {
+  # variables are defined in bundle common g
+  classes:
+      # Policy Server is a machine which delivers promises
+      "policy_server" expression => strcmp("root","${g.uuid}");
+      # Root Server is the top policy server machine
+      "root_server" expression => strcmp("root","${g.uuid}");
 
-  vars:
-    windows.!cygwin::
-      "rudder_base" string  => "${sys.winprogdir}\Rudder";
-      "uuid_file" string => "${rudder_base}\etc\uuid.hive";
 
-    !windows.!android::
-      "rudder_base" string => "/opt/rudder";
-      "uuid_file" string => "${rudder_base}/etc/uuid.hive";
+      # on changes only, we only reports on repaired or error message
+      "changes_only"     expression => strcmp("full-compliance", "changes-only");
+      "reports_disabled" expression => strcmp("full-compliance", "reports-disabled");
 
-    android::
-      "rudder_base" string => "/data/rudder";
-      "uuid_file" string => "${rudder_base}/etc/uuid.hive";
-
-    any::
-      "uuid" string => readfile("${uuid_file}", 60);
+      # full compliance is the default mode
+      "full_compliance" not => "changes_only|reports_disabled";
 }
 
 body common control
 {
-
     any::
         output_prefix => "rudder";
         
         protocol_version   => "classic";
 
         inputs => {
-  @{va.inputs_list}
-    };
+          @{va.inputs_list},
+          "rudder_system_directives.cf",
+          "rudder_directives.cf"
+        };
 
-    policy_server::
+        bundlesequence => { 
+          @{va.common_bs},
+          @{va.specific_bs},
+          rudder_system_directives,
+          rudder_directives,
+          @{va.end}
+        };
 
-        bundlesequence => { @{va.bs}, "install_rsyslogd", "propagatePromises", "root_component_check", "fetchFusionTools", "virtualMachines", "doInventory", "sendInventoryToCmdb", @{va.end}};
-
-    !policy_server::
-
-        bundlesequence => { @{va.bs} , "fetchFusionTools", "virtualMachines", "doInventory" , @{va.end}};
-
+    android::
+        # if native syslog is not available, use cfengine implementation (eg: android)
+        syslog_host => "${server_info.cfserved}";
+        syslog_port => "${install_rsyslogd.rsyslog_port[2]}";
 }
 
 bundle common va
 {
-
   vars:
+
+      "end" slist => {
+        "endExecution"
+      };
+
+      # all other inputs are loaded by body file control in rudder_system_directives.cf and rudder_directives.cf
+      "inputs_list" slist => { @{ncf_inputs} };
+
+      "common_bs" slist => {
+        "startup",
+        "check_disable_agent",
+        "clean_red_button",
+        "update",
+        "configuration",
+        "set_red_button",
+        "internal_security",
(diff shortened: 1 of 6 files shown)


initial-promises/node-server/promises.cf
```diff
@@ -142,7 +224,7 @@ bundle agent check_disable_agent
       # Only define this class when we're ready to die - this is a special class name in "abortclasses"
       "should_not_continue"
         expression => "should_disable_agent",
-        ifvarclass => "abort_report_done",
+        ifvarclass => "abort_report_done";
```
jooooooon: This won't work, due to the following line

peckpeck:
PR rebased

techniques/system/common/1.0/promises.st
```diff
@@ -62,43 +91,47 @@ bundle common va
...
-        "nxlog_enable",
-        "check_zypper",
-        "check_uuid",
-        "get_environment_variables",
-        "properties",
-        "configuration"
       };
 
+&if(INITIAL)&
+      # initial promises server specific inputs cannot be loaded from rudder_system_directives.cf
```
jooooooon: Why not?

techniques/system/common/1.0/promises.st
```diff
@@ -62,43 +91,47 @@ bundle common va
...
         "garbage_collection",
         "setup_cronjob",
         "check_cron_daemon",
-        "check_binaries_freshness",
+&if(INITIAL)&
+        "fetchFusionTools",
+        "virtualMachines",
+        "doInventory",
+&else&
+        "configure_rudder_reporting_system",
```
jooooooon: This should be in a "!INITIAL" block

techniques/system/common/1.0/promises.st
peckpeck: Because they are generated there in a String Template variable that is dynamic in the webapp but static in initial promises. So we need a condition to choose them at runtime with a class here and not at generation time.

techniques/system/common/1.0/promises.st
peckpeck: Doesn't the &else& work ?

jooooooon: Oh. I just hadn't read that *at all*. Sorry... :S

techniques/system/common/1.0/promises.st
jooooooon: Ah, I see.

techniques/system/common/1.0/promises.st
```diff
@@ -47,11 +70,17 @@ bundle common va
 {
   vars:
 
-      "end" slist => { "restart_services", "endExecution" };
+      "end" slist => {
+&if(!INITIAL)&
+        "restart_services", 
+&endif&
```
jooooooon: OK, this is used to restart syslog services, so initial promises don't need it. Got it.

techniques/system/common/1.0/promises.st
```diff
@@ -62,43 +91,47 @@ bundle common va
...
+      # initial promises server specific inputs cannot be loaded from rudder_system_directives.cf
+      "inputs_list" slist => { @{ncf_inputs}, &RUDDER_SERVER_SYSTEM_DIRECTIVES_INPUTS& };
+&endif&
+
+    !policy_server::
+
+      # This bundle is duplicated to avoid having empty lists and cfnull in bundlesequence
+      "specific_bs" slist => { "check_binaries_freshness" }
+
+      "inputs_list" slist => { @{ncf_inputs} };
```
jooooooon: This seems to be an identical line to the one above that is without a class context (so the same on all server)

jooooooon: I think it should be removed

techniques/system/common/1.0/promises.st
```diff
@@ -21,21 +21,44 @@
...
+        bundlesequence => { 
+          @{va.common_bs},
+          @{va.specific_bs},
+          rudder_system_directives,
+          rudder_directives,
+          @{va.end}
+        };
 
     android::
         # if native syslog is not available, use cfengine implementation (eg: android)
```
jooooooon: This block didn't use to be in initial promises, and since initial promises shouldn't send reports, I think that it still shouldn't be, because I believe this causes CFEngine to send messages to syslog. Could you surround it with if(INITIAL)?

techniques/system/common/1.0/promises.st
```diff
@@ -47,11 +70,17 @@ bundle common va
...
+      "end" slist => {
+&if(!INITIAL)&
+        "restart_services", 
+&endif&
+        "endExecution"
+      };
 
-    policy_server::
+      # all other inputs are loaded by body file control in rudder_system_directives.cf and rudder_directives.cf
+      "inputs_list" slist => { @{ncf_inputs} };
```
jooooooon: OK, so let's remove it here

variables.json
```diff
@@ -18,5 +18,17 @@
...
+  "RUDDER_REPORT_MODE": "full-compliance",
+  "NODEROLE": "# This node doesn't have any specific role",
+  "RUDDER_DIRECTIVES_INPUTS": "",
+  "RUDDER_DIRECTIVES_SEQUENCE": "",
+  "RUDDER_SYSTEM_DIRECTIVES_INPUTS": "\"common/1.0/cf-served.cf\",\"common/1.0/rudder-stdlib.cf\",\"common/1.0/rudder-stdlib-core.cf\",\"common/1.0/rudder_lib.cf\",\"common/1.0/cron_setup.cf\",\"common/1.0/internal_security.cf\",\"common/1.0/site.cf\",\"common/1.0/update.cf\",\"inventory/1.0/fetchFusionTools.cf\",\"inventory/1.0/virtualMachines.cf\",\"inventory/1.0/fusionAgent.cf\"",
+  "RUDDER_SERVER_SYSTEM_DIRECTIVES_INPUTS": "\"distributePolicy/1.0/propagatePromises.cf\",\"distributePolicy/1.0/rsyslogConf.cf\",\"server-roles/1.0/service-check.cf\",\"server-roles/1.0/postgres-check.cf\",\"server-roles/1.0/logrotate-check.cf\", \"server-roles/1.0/servers-by-role.cf\", \"server-roles/1.0/integrity-check.cf\",\"server-roles/1.0/technique-reload.cf\",\"server-roles/1.0/network-check.cf\",\"server-roles/1.0/alive-check.cf\",\"server-roles/1.0/password-check.cf\", \"server-roles/1.0/component-check.cf\"",
```
jooooooon: Let's remove this as a ST variable and use the value directly

peckpeck:
Commit modified

PR #1005, opened 2016-08-04
Fixes #8816: Misleading comment in rudder cron

peckpeck:
https://www.rudder-project.org/redmine/issues/8816


amousset:
Shouldn't we also remove this from initial promises ?

peckpeck:
OK, merging this PR

PR #1008, opened 2016-08-12
Fixes #8841: Typo in system techniques reports

amousset:
https://www.rudder-project.org/redmine/issues/8841


amousset:
OK, merging this PR

PR #918, opened 2016-05-09
Fixes #7621: cf-promises can hang for about 30 seconds looking up default server name \"rudder\"

jooooooon:
https://www.rudder-project.org/redmine/issues/7621

Replacing previous PR: https://github.com/Normation/rudder-techniques/pull/819

PR #1012, opened 2016-08-26
Fixes #8900:  cf-promises can hang for about 30 seconds looking up default server name "rudder" - initial promises

amousset:
https://www.rudder-project.org/redmine/issues/8900


amousset:
Same as #918 for initial promises.

amousset:
OK, merging this PR

PR #1015, opened 2016-08-30
Fixes #8825: crontab frequency does not support reserved string @reboot

amousset:
https://www.rudder-project.org/redmine/issues/8825


amousset:
Commit modified

jooooooon:
I don't think this can work. This would create a file in a directory called /etc/cron.reboot. I don't think that is a thing.

The aim here is to use a specific time setting of "@reboot" (or @yearly, @annually) as the time specifier in a cron.d file (in place of something like "0 0 1 * *").

Therefore, the easiest path to implementation is just to adapt the regex for "specify manually" to allow for these three values: something like `(<existing regex>|@(yearly|annually|reboot))`

amousset:
Commit modified

amousset:
Commit modified

techniques/systemSettings/systemManagement/cronManagement/3.0/metadata.xml
```diff
@@ -14,6 +14,7 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 <TECHNIQUE name="Cron daemon configuration">
+  <DEPRECATED>This technique version has been superseded by a new version. It will no longer be available in the next stable version of Rudder. Please upgrade to the latest version.</DEPRECATED>
```
jooooooon: Hmmm, I don't think we can reasonably *deprecate* versions of techniques in the same major version as they were supported in! This should be added in master branch instead.

jooooooon:
Looks good to me apart from deprecation!

techniques/systemSettings/systemManagement/cronManagement/3.0/metadata.xml
amousset: It was done by rudder-dev in the init commit, fixing.

amousset:
Commit modified

amousset:
OK, merging this PR

PR #1020, opened 2016-09-07
Fixes #8954: Package version match in "Package management for RHEL / CentOS / SuSE / RPM systems"

amousset:
https://www.rudder-project.org/redmine/issues/8954


amousset:
OK, merging this PR

PR #1025, opened 2016-09-08
Fixes #8984: rudder-rsyslog-root.conf wring name in initial promises

peckpeck:
https://www.rudder-project.org/redmine/issues/8984


peckpeck:
Commit modified

PR #1029, opened 2016-09-12
Fixes #9003: Update modules from ncf internal in Rudder's failsafe

amousset:
https://www.rudder-project.org/redmine/issues/9003

### DIFF
--- a/techniques/system/common/1.0/update.st
+++ b/techniques/system/common/1.0/update.st
@@ -237,6 +237,16 @@ bundle agent update_action
         classes           => success("rudder_ncf_local_updated", "rudder_ncf_local_update_error", "rudder_ncf_local_updated_ok"),
         comment           => "Update the local Rudder ncf instance";
 
+      # Copy ncf modules
+      "${sys.workdir}/modules/"
+        copy_from         => copy_digest_without_perms("${g.rudder_ncf}/common/10_ncf_internals/modules/"),
+        depth_search      => recurse_ignore("inf", @{g.excludedreps}),
+        perms             => u_mog("755", "root", "0"),
+        action            => immediate,
+        move_obstructions => "true",
+        classes           => success("rudder_ncf_common_modules_updated", "rudder_ncf_common_modules_update_error", "rudder_ncf_common_modules_updated_ok"),
+        comment           => "Update the common Rudder ncf modules";
+
     !root_server::
       "${client_inputs}/${rudder_promises_timestamp_tmp}"
         copy_from         => remote_unsecured_without_perms("${server_info.cfserved}","${server_inputs}/${rudder_promises_timestamp}"),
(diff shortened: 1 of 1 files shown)


amousset:
WIP, do not merge.

amousset:
* Should we check at each run or only when ncf/common was repaired?
* Should we handle local modules ? In this case, we need to handle the case were the modules folder does not exist.

ncharles:
I think we should use the same method for update, so copy the module only if the hash of the ncf was updated
And we need to create the modules folder if it doesn't exists

jooooooon:
I don't think this should be done here at all. As explained in the ncf source code in Normation/ncf#427, copying these files is required for the new package_* methods in ncf to work. Therefore, ncf should do it itself if necessary.

This may require adding a "ncf initialisation" bundle that is called by Rudder, very early in the bundlesequence, if we don't already have one, but that's fine, since it's not a hard change, and I'm sure it will come in for other reasons.

Also, I don't like this being in update.cf because this isn't really a copy from the server, it's just a local copy, so that doesn't need to be done at a specific point in time - unless of course these files are loaded on cf-agent startup rather than when a packages: promise is called?

ncharles:
Updating parts of files of workdir outside the update will be pretty surprising - it is a big change from the way files in this folder are handled in both Rudder and CFEngine usually, and it will surprise users trying to understand why content of modules folder change outside of the update

jooooooon:
> Updating parts of files of workdir outside the update will be pretty surprising - it is a big change from the way files in this folder are handled in both Rudder and CFEngine usually, and it will surprise users trying to understand why content of modules folder change outside of the update

Lots of parts of the workdir are changed outside update - outputs are written at the end of a run, package lists are written during the run, the whole state directory is updated during the run, the ncf.conf.cache file is generated during the run, ncf's expected promises files are generated during the run, ...

Let's not get caught up in internal details here please - this is a ncf requirement, it belongs in ncf.

ncharles:
`Let's not get caught up in internal details here please - this is a ncf requirement, it belongs in ncf.`

Same as propagating in some way to the nodes the folders 10, 20, 30, etc that belong in ncf - yet they are updated with rudder

ncharles:
I doubled checked with CFEngine vanilla implementation: by default, they do not purge the modules folder when copying it, so we can populate it from outside an update.cf (even if it will be surprising for user, it won't be non-convergent)

amousset:
Updating the modules from outside the update process is indeed against the "masterfiles way". However, if we want ncf to stay, as much as possible (because files loading is a big exception), an independent framework, used by Rudder policies, it is probably a better idea to use a defined and stable API like an init bundle that to interfere with ncf internals (even if in this case modules are both internal for ncf and global for the agent).

We can add this call to ncf init in system techniques (but won't be able to call it from update.cf as it would require to load ncf). I think we also need to add a report for this step, as it is important and does not match any of current reports.

fanf:
It seems to have a consensus on that way of doing the update. So I'm closing that PR as the code will live in ncf.

PR #1034, opened 2016-09-13
Fixes #9035: Typo in common technique reports

amousset:
https://www.rudder-project.org/redmine/issues/9035


amousset:
OK, merging this PR

PR #1035, opened 2016-09-14
Fixes #8884: Change .pgpass configuration to use the right values

amousset:
https://www.rudder-project.org/redmine/issues/8884


amousset:
OK, merging this PR

PR #1037, opened 2016-09-15
Fixes #9060: Specified action bodies override the dry-run policy - Techniques part

amousset:
https://www.rudder-project.org/redmine/issues/9060


amousset:
I did not add it to failsafe and update as it does not really make sense, and we probably want to prevent a broken dry-run to prevent update of policies.

ncharles:
That seems the right course of action, thank you !

PR #1039, opened 2016-09-20
Fixes #9041: Change system promises to call list-compatible-inputs with Rudder version

amousset:
https://www.rudder-project.org/redmine/issues/9041

### DIFF
--- a/techniques/system/common/1.0/promises.st
+++ b/techniques/system/common/1.0/promises.st
@@ -110,6 +110,7 @@ bundle common va
       "rudder_tools"   string => "${rudder_var}/tools";
       "shell_type"     string => "useshell";
       "list_compatible_inputs" string => "NCF_CACHE_PATH=${sys.workdir}/state /bin/sh ${ncf_path}/common/10_ncf_internals/list-compatible-inputs";
+      "rudder_version_file" string => "${rudder_base}/etc/rudder-version";
 
 
     windows::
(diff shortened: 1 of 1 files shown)


amousset:
Commit modified

amousset:
Commit modified

techniques/system/common/1.0/promises.st
```diff
@@ -119,21 +120,25 @@ bundle common va
...
-
-      "path_ncf_local_inputs_10" slist => splitstring(execresult("${list_compatible_inputs} ${sys.cf_version} '${ncf_path}' local/10_ncf_internals", "${shell_type}"), "\n", 10000);
-      "path_ncf_local_inputs_20" slist => splitstring(execresult("${list_compatible_inputs} ${sys.cf_version} '${ncf_path}' local/20_cfe_basics", "${shell_type}"), "\n", 10000);
-      "path_ncf_local_inputs_30" slist => splitstring(execresult("${list_compatible_inputs} ${sys.cf_version} '${ncf_path}' local/30_generic_methods", "${shell_type}"), "\n", 10000);
-      "path_ncf_local_inputs_40" slist => splitstring(execresult("${list_compatible_inputs} ${sys.cf_version} '${ncf_path}' local/40_it_ops_knowledge", "${shell_type}"), "\n", 10000);
-      "path_ncf_local_inputs_50" slist => splitstring(execresult("${list_compatible_inputs} ${sys.cf_version} '${ncf_path}' local/50_techniques", "${shell_type}"), "\n", 10000);
-      "path_ncf_local_inputs_60" slist => splitstring(execresult("${list_compatible_inputs} ${sys.cf_version} '${ncf_path}' local/60_services", "${shell_type}"), "\n", 10000);
+      "version_data" data => readjson("${rudder_version_file}", 4000);
```
jooooooon: I think this will fail on a node that has not got a recent rudder-agent (so no rudder_version_file on the filesystem). This can happen if ncf is upgraded on the rudder server, without an agent upgrade - a common scenario.

I think we should protect against this file not existing. (check it exists before trying to read it)

peckpeck:
We will use a capability based filter instead

PR #1041, opened 2016-09-22
Fixes #9041: Change system promises to call list-compatible-inputs with capabilities file

peckpeck:
https://www.rudder-project.org/redmine/issues/9041


techniques/system/common/1.0/promises.st
```diff
@@ -108,7 +108,9 @@ bundle common va
       "ncf_path"       string => "${rudder_var}/ncf";
       "rudder_tools"   string => "${rudder_var}/tools";
       "shell_type"     string => "useshell";
+      "capability_file"        string => "/opt/rudder/etc/agent-capabilities"; # no path variable defined yet
```
amousset: We should improve this in 4.0, and move path definition here.

peckpeck: yes

PR #1043, opened 2016-09-23
Fixes #9113: Techniques use \{sys.workdir} for paths of ressources fi…

ncharles:
…les (ex: properties), which prevent their validation by the Web Interface during policies generation


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1049, opened 2016-10-05
Fixes #9228: Add ncf standard abort class to rudder abort classes

amousset:
https://www.rudder-project.org/redmine/issues/9228


amousset:
OK, merging this PR

PR #1052, opened 2016-10-08
Fixes #6353: Secure agent/server communication

ncharles:
https://www.rudder-project.org/redmine/issues/6363

### DIFF
--- a/techniques/system/common/1.0/failsafe.st
+++ b/techniques/system/common/1.0/failsafe.st
@@ -27,7 +27,7 @@ body common control
         inputs             => { "common/1.0/update.cf", "common/1.0/rudder-stdlib-core.cf" };
         output_prefix      => "rudder";
 
-        protocol_version   => "classic";
+        protocol_version   => "2";
 }
 
 body agent control {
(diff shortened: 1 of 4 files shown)


peckpeck:
I don't like this pull request being in the issue meant to aggregate all security related improvement.
This hould be in https://www.rudder-project.org/redmine/issues/6353 instead.

techniques/system/common/1.0/failsafe.st
```diff
@@ -27,7 +27,7 @@ body common control
         inputs             => { "common/1.0/update.cf", "common/1.0/rudder-stdlib-core.cf" };
         output_prefix      => "rudder";
 
-        protocol_version   => "classic";
+        protocol_version   => "2";
```
peckpeck: This won't work on older servers. We need a transition method.
- easy path: leave classic until old server are not supported anymore
- bad path: ask the user
- hard path: test both cases starting with the secure one allowing an override to forbid the old one

peckpeck: Alternative: 
- old protocol in initial promises
- new protocol in system technique

ncharles:
ha, i'm retargeting the PR, sorry (it's a typo)

ncharles:
commit renamed

techniques/system/common/1.0/failsafe.st
ncharles: An agent protocol v2 can connect to a server protocol v1 (i just tried it)

PR #1058, opened 2016-10-17
Fixes #9359: Agent capabilities should be per agent, and not per node

peckpeck:
https://www.rudder-project.org/redmine/issues/9359


ncharles:
Thank you

PR #1061, opened 2016-10-19
Fixes #9449: Syntax error in promises.cf

amousset:
https://www.rudder-project.org/redmine/issues/9449


amousset:
Commit modified

PR #1070, opened 2016-10-21
Fixes #9216: Error message when restarting rudder-agent during relay install

ncharles:
https://www.rudder-project.org/redmine/issues/9216

### DIFF
--- a/techniques/system/common/1.0/cf-served.st
+++ b/techniques/system/common/1.0/cf-served.st
@@ -105,17 +105,6 @@ bundle server access_rules
       "${sys.workdir}\bin\cf-agent.exe"
         admit   => { host2ip("${server_info.cfserved}"), string_downcase(escape("${server_info.cfserved}")) };
 
-  &if(NOVA)&
-    any::
-      "${query_types}"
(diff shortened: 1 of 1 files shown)


techniques/system/common/1.0/cf-served.st
```diff
@@ -105,17 +105,6 @@ bundle server access_rules
...
 
-  &if(NOVA)&
-    any::
-      "${query_types}"
-        handle => "server_access_grant_${query_types}_for_hosts",
-        comment => "Grant ${query_types} reporting query for the hub on the hosts",
-        resource_type => "query",
-        report_data_select => rudder_data_select_host,
-        admit => { ${def.policy_server}, @{sys.ip_addresses} };
-  &endif&
```
amousset: Why was this code here in the first place?

amousset: ping @ncharles

ncharles: This is for interconecting rudder/mission-portal

ncharles: this is a duplicate of l69

PR #1080, opened 2016-11-08
Fixes #9626: The locks are removed every 10 minutes by check-rudder-agent when there is no promise update

amousset:
https://www.rudder-project.org/redmine/issues/9626


ncharles:
Looks good. But this code is complex

PR #1083, opened 2016-11-23
Fixes #9716: A cfengine enterprise agent fails to connect to rudder server in cfengine 3.7, because there is no agentallconnects

ncharles:
https://www.rudder-project.org/redmine/issues/9716


ncharles:
Commit modified

PR #1085, opened 2016-11-24
Fixes #9710: Generate a file in relays promises that contain the map of nodes and relays below it

peckpeck:
https://www.rudder-project.org/redmine/issues/9710


amousset:
Looks good, you can self merge when variable names are confirmed.

amousset:
In fact we should wait for the variables to be available in the webapp to avoid breaking generation.

PR #1086, opened 2016-11-24
Fixes #9709: Add the path of files shared between nodes to cf-serverd configuration

peckpeck:
https://www.rudder-project.org/redmine/issues/9709


peckpeck:
Commit modified

peckpeck:
Commit modified

PR #1091, opened 2016-12-02
Fixes #9761: Allow to only install a specific version in apt and rpm techniques.

amousset:
https://www.rudder-project.org/redmine/issues/9761


ncharles:
\o/

amousset:
OK, merging this PR

PR #1093, opened 2016-12-15
Fix #9794: Malformed .gz inventory breaks upload process

armeniaca:


tools/send-clean.sh
```diff
@@ -67,6 +67,16 @@ fi
 if [ "${BASENAME##*.}" = "gz" ]
 then
   ${GZIP_BINARY} --force --quiet --decompress ${FILENAME}
+  UNZIP_COMMAND_RET=$?
+  if [ $UNZIP_COMMAND_RET -ne 0 ]
+  then
+    echo "ERROR: failed to decompress inventory in ${FILENAME} az gzip, putting it in the failed directory"
+    mv "${FILENAME}" "${FAILEDDIR}/`basename ${FILENAME}`-$(date --rfc-3339=date)"
```
peckpeck: why do you mix `` and $() ?

tools/send-clean.sh
```diff
@@ -67,6 +67,16 @@ fi
...
+    mv "${FILENAME_SIGN}" "${FAILEDDIR}/`basename ${FILENAME_SIGN}`-$(date --rfc-3339=date)"
```
peckpeck: There is not always a signature file, you should avoid errors here with a [ -f ] &&

armeniaca: same as above. will rework that part of code without using your logic, but as I see well suited.

peckpeck:
Sorry for the original mixed style and thank you for the PR.

PR #1096, opened 2016-12-27
Fixes #9822: Job is not run for Technique JobScheduler in Rudder

ncharles:
https://www.rudder-project.org/redmine/issues/9822


ncharles:
ping @amousset

PR #1097, opened 2016-12-29
Fix typo

aanriot:
Hello,

Here is quick typo fix.

Should it be also applied to branches/rudder/3.2 and branches/rudder/4.0 ?

Thanks.


amousset:
Hello @aanriot,

> Should it be also applied to branches/rudder/3.2 and branches/rudder/4.0 ?

Yes, pull requests should always target the lowest branch having the issue (and the fix will be upmerged automatically when possible), so it should be targeted to `branches/rudder/3.2`.

aanriot:
Hello @amousset,

I was wondering if it was corresponding to your packaging policy etc. I will reopen another PR then.

Thanks.

PR #1100, opened 2017-01-05
Fixes #9776: Logs from 4.1 agent are not sent to the server

amousset:
https://www.rudder-project.org/redmine/issues/9776


ncharles:
This fix means that a Rudder-agent 4.1 will not send reports to a Rudder-server 3.1. Is it expected ?

amousset:
Reverted the change in https://github.com/Normation/rudder-packages/pull/1169

amousset:
This should be done on 3.1

PR #1106, opened 2017-01-12
Fixes #9975: We sometimes write a wrong require statement in apache 2.4 conf

amousset:
https://www.rudder-project.org/redmine/issues/9975


amousset:
Commit modified

PR #1110, opened 2017-01-20
Fixes #9979: Simple relay tries to configure rsyslog-pgsql

amousset:
https://www.rudder-project.org/redmine/issues/9979


amousset:
Before 4.0, the rsyslog-pgsql package was installed on all relays.

PR #1111, opened 2017-01-27
Fixes #10081: Use rudder agent run as cfruncommand

amousset:
https://www.rudder-project.org/redmine/issues/10081

### DIFF
--- a/techniques/system/common/1.0/cf-served.st
+++ b/techniques/system/common/1.0/cf-served.st
@@ -114,7 +114,7 @@ bundle server access_rules
         admit   => { host2ip("${server_info.cfserved}"), string_downcase(escape("${server_info.cfserved}")) };
 
     !windows::
-      "${def.cf_runagent_shell}"
+      "${g.rudder_command}" comment => "Grant access to rudder agent command for cfruncommand",
         admit => { host2ip("${server_info.cfserved}"), string_downcase(escape("${server_info.cfserved}")) };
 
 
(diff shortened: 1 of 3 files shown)


techniques/system/common/1.0/cf-served.st
```diff
@@ -246,8 +246,9 @@ body server control
         # last single quote in cfruncommand is left open, so that
         # arguments (like -K and --remote-bundles) are properly appended.
         cfruncommand => "${def.cf_runagent_shell} -c \'
-                            ${sys.cf_agent} -I -D cfruncommand -f failsafe.cf  \&\&
-                            ${sys.cf_agent} -I -D cfruncommand";
+                            ${g.rudder_command} agent update &&
```
ncharles: you need to backslash the &

ncharles: and you will need an access rule for g.rudder_command, like
`"$(g.rudder_command)"
      comment => "Grant access to rudder agent command for cfruncommand",
      admit => { ${def.policy_server} };`

amousset:
Commit modified

amousset:
Commit modified

amousset:
@peckpeck @ncharles Made some important changes:

* Avoid having to authorize to use a shell (that went against the idea of limiting authorized commands) by calling `rudder agent run -u` directly
* Stop conditioning agent execution to policy update success. Do you know what was the point of that?

techniques/system/common/1.0/cf-served.st
```diff
@@ -114,7 +114,7 @@ bundle server access_rules
         admit   => { host2ip("${server_info.cfserved}"), string_downcase(escape("${server_info.cfserved}")) };
 
     !windows::
-      "${def.cf_runagent_shell}"
+      "${g.rudder_command}" comment => "Grant access to rudder agent command for cfruncommand",
```
peckpeck: So we won't need the shll anywhere ?

peckpeck:
I think it was because if download fails, promises are probably not up to date.
But running rudder anyway is closer to hat we expect

amousset:
Commit modified

amousset:
ping @peckpeck

amousset:
Commit modified

PR #1124, opened 2017-03-08
Fixes #10312: "Download files from the shared folder" behaving badly?

amousset:
https://www.rudder-project.org/redmine/issues/10312

### DIFF
--- a/techniques/fileDistribution/copyGitFile/2.0/copyFileFromSharedFolder.st
+++ b/techniques/fileDistribution/copyGitFile/2.0/copyFileFromSharedFolder.st
@@ -80,6 +80,11 @@ bundle agent download_from_shared_folder
       string     => "0",
       ifvarclass => "!enable_suid_${index}.!enable_sgid_${index}";
 
+  defaults:
+      # Add a trailing slash to force considering destination as a directory and avoid replacing it with a file
+      "copyfile[${index}][destination]" string => "${copyfile[${index}][destination]}/", if_match_regex => ".*[^\/]",
+                                    ifvarclass => "pass1.!is_file_${index}";
+
   classes:
       "is_valid" not => strcmp("&SHARED_FILES_FOLDER&", "");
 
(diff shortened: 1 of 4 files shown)


amousset:
First version of a fix, current status:

* Prevents replacing a directory by a file, gives an error
* Leaves a `.cfnew` on destination

We could try to be more clever, and use the file name from source as destination, in addition to the existing directory path.

techniques/fileDistribution/copyGitFile/2.0/copyFileFromSharedFolder.st
```diff
@@ -80,6 +80,11 @@ bundle agent download_from_shared_folder
       string     => "0",
       ifvarclass => "!enable_suid_${index}.!enable_sgid_${index}";
 
+  defaults:
+      # Add a trailing slash to force considering destination as a directory and avoid replacing it with a file
+      "copyfile[${index}][destination]" string => "${copyfile[${index}][destination]}/", if_match_regex => ".*[^\/]",
+                                    ifvarclass => "pass1.!is_file_${index}";
```
ncharles: can you define defaults on pass1 ?

techniques/fileDistribution/copyGitFile/2.0/metadata.xml
```diff
@@ -81,7 +81,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           </SECTION>
       <INPUT>
         <NAME>COPYFILE_DESTINATION</NAME>
-        <DESCRIPTION>Destination of the file</DESCRIPTION>
+        <DESCRIPTION>Full path of destination file or directory</DESCRIPTION>
       <LONGDESCRIPTION>This is the absolute path of the file/folder on the managed node</LONGDESCRIPTION>
```
ncharles: i think you can add more details in the long description, like an example
"for instance, /etc/cron.d/0hourly", or whatever loofs relevant

techniques/fileDistribution/copyGitFile/2.0/copyFileFromSharedFolder.st
ncharles: i'm pretty sure I tested defaults before, and that they failed if not evaluated at first pass

amousset: I tested it, and it seemed to work well (but it was on Rudder 4.1, I will check with 3.1)

amousset: Tested on 3.1 too.

ncharles:
Thank you !

PR #1128, opened 2017-03-13
Fixes #10397: Missing "default-directive-names.conf" example file

fanf:
https://www.rudder-project.org/redmine/issues/10397


fanf:
PR rebased

techniques/default-directive-names.conf
```diff
@@ -0,0 +1,22 @@
+
+#
+# This file contains the default name that a directive gets in Rudder UI creation pop-up.
```
amousset: There are no pop-ups anymore :)

techniques/default-directive-names.conf
```diff
@@ -0,0 +1,22 @@
...
+# available "TechniqueName" from this file. If neither key is available, the
+# pop-up will use the actual Technique name as default.
+# 
+# Don't forget to commit the file in git (git add default-directive-names.conf; 
+# git commit -m "change new directive name) and reload technique library
+# (rudder server reload-techniques) to have modifications seen by Rudder. 
+#
+
+# Default pattern for new directive from "userManagement" technique:
+userManagement=User: Role <name>: <login>
```
amousset: Shouldn't this be commented ?

fanf: I was wondering. It is inocuitious but make the feature more discoverable (why default name is that for User?). But perhaps the correct thing to do would be to provide a default set of name for core techniques, and not only user (and so for now, comment that out).

fanf:
PR rebased

techniques/default-directive-names.conf
amousset: I think we should stay consistent, at least over core techniques, and provide better names.

PR #1132, opened 2017-03-31
Fixes #10473: Add tests for manageKeyValueFile

peckpeck:
https://www.rudder-project.org/redmine/issues/10473


peckpeck:
When there are many inits, all of them are run.
We don't do a cross product of everything in the metadata, we do all inits, then all directives, then all checks.
The goal was to make it easy to factor them.

ncharles:
Ok - makes sense

PR #1135, opened 2017-04-21
Fixes #10641: Unecessary call to inventory code to check its version during policy validation

ncharles:
https://www.rudder-project.org/redmine/issues/10641


peckpeck:
💯

PR #1141, opened 2017-04-27
Fixes #8436: Getting server uuid fails on agent with old openssl

peckpeck:
https://www.rudder-project.org/redmine/issues/8436

### DIFF
--- a/initial-promises/node-server/inventory/1.0/fusionAgent.cf
+++ b/initial-promises/node-server/inventory/1.0/fusionAgent.cf
@@ -89,10 +89,14 @@ bundle agent doInventory_always
 bundle agent download_server_uuid
 {
   vars:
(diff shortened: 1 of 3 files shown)


peckpeck:
ping @amousset

peckpeck:
Very old pr since june 12 @amousset

amousset:
Do we know what is the real cause of the issue? We know it works on SLES11 at least in some cases. This PR seems to reallow SSLv2/3 which does not look like a good solution if not strictly necessary.

peckpeck:
It does, on old sles.
We should ask the bug reporter if we want more information on the condition, but it seems blurry

peckpeck:
An up to date sles should have an up to date openssl.
Lets' package openssl for very old os instead.

PR #1153, opened 2017-06-14
Fixes #9283: Use different package_name_convention in rpm package_method body when there is a defined version

amousset:
https://www.rudder-project.org/redmine/issues/9283


amousset:
Still requires some testing.

amousset:
@ncharles Tested on RHEL6.

PR #1155, opened 2017-06-19
Fixes #10949: Agent output contains error because of new syslog service checks

amousset:
https://www.rudder-project.org/redmine/issues/10949

### DIFF
--- a/techniques/system/common/1.0/promises.st
+++ b/techniques/system/common/1.0/promises.st
@@ -538,9 +538,9 @@ bundle agent check_log_system
   classes:
 
     !android::
-      "rsyslogd" expression => fileexists("${rsyslog_conffile}");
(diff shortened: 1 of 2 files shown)


techniques/system/common/1.0/promises.st
```diff
@@ -538,9 +538,9 @@ bundle agent check_log_system
   classes:
 
     !android::
-      "rsyslogd" expression => fileexists("${rsyslog_conffile}");
-      "syslogng" expression => fileexists("${syslogng_conffile}");
-      "syslogd"  expression => fileexists("${syslog_conffile}");
+      "rsyslogd_conffile_present" expression => fileexists("${rsyslog_conffile}"), scope => "namespace";
```
amousset: I prefer avoiding too generic classes in default namespace, but this can be discussed.

PR #1159, opened 2017-06-22
Fixes #10987: No logs when (r)?syslog(-ng)? is restarted in 4.1

ncharles:
https://www.rudder-project.org/redmine/issues/10987

### DIFF
--- a/techniques/system/common/1.0/restart-services.st
+++ b/techniques/system/common/1.0/restart-services.st
@@ -78,4 +78,9 @@ bundle agent restart_services
     # We have a problem only if all 3 have an error (otherwise at least one is running)
     service_ensure_running_syslog_error.service_ensure_running_syslogng_error.service_ensure_running_rsyslog_error::
       "any"  usebundle => rudder_common_report("Common", "result_error", "&TRACKINGKEY&", "Log system for reports", "None", "Could not start the logging system");
+
+    # Is syslog service had been restarted, we need to log it as well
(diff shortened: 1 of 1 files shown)


techniques/system/common/1.0/restart-services.st
```diff
@@ -78,4 +78,9 @@ bundle agent restart_services
     # We have a problem only if all 3 have an error (otherwise at least one is running)
     service_ensure_running_syslog_error.service_ensure_running_syslogng_error.service_ensure_running_rsyslog_error::
       "any"  usebundle => rudder_common_report("Common", "result_error", "&TRACKINGKEY&", "Log system for reports", "None", "Could not start the logging system");
+
+    # Is syslog service had been restarted, we need to log it as well
+    (service_ensure_running_rsyslog_repaired.!service_restart_rsyslog_repaired)|(service_ensure_running_syslog_repaired.!service_restart_syslog_repaired)|(service_ensure_running_syslogng_repaired.!service_restart_syslog_ng_repaired)::
```
amousset: I don't think we need to use the result of service restart, it would only hide a real issue (= the service was restarted but not running).

ncharles: we already have reports based on service_restart, and restart is when we changed the configuration

amousset: yes but we are only doing log_repaired, so we are not limited to one?

amousset:
Replaced by https://github.com/Normation/rudder-techniques/pull/1160/files

PR #1161, opened 2017-06-22
Fixes #10997: Missing template distributePolicy/1.0,apache_acl during generation

ncharles:
https://www.rudder-project.org/redmine/issues/10997


ncharles:
don't merge, i have an error at execution

PR #1162, opened 2017-06-22
Fixes #10998: Apache acl are not correcly updated in 4.2

ncharles:
https://www.rudder-project.org/redmine/issues/10998


peckpeck:
I don't understand the fix since this is the default outpath

amousset:
Sadly, the default outpath adds `.cf`.

peckpeck:
Then we should re-add nodeslist.json

PR #1166, opened 2017-06-27
Fixes #11044: Wrong stderr / stdout redirection of logs from rudder agent check in 4.1

ncharles:
https://www.rudder-project.org/redmine/issues/11044


amousset:
If this depends on the agent that executes it, we cannot fix it on the server as we don't know which agent versions will be used.

ncharles:
no, it depends on the server - in 3.1 it's a .st file, here it's a copied file (so no .st, and no need to escape the &)

peckpeck:
Indeed it depends on the server version

PR #1170, opened 2017-07-06
Fixes #11106: 3787 is still present in 4.0+ versions of the technique

amousset:
https://www.rudder-project.org/redmine/issues/11106


amousset:
OK, merging this PR

PR #1191, opened 2017-09-13
Fixes #11087: File content (key/value format) technique allows white space before separator but not after it

ncharles:
https://www.rudder-project.org/redmine/issues/11087

### DIFF
--- a/techniques/fileDistribution/manageKeyValueFile/1.0/metadata.xml
+++ b/techniques/fileDistribution/manageKeyValueFile/1.0/metadata.xml
@@ -16,7 +16,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 <!-- This technique lets user define a key-value in a file -->
 <TECHNIQUE name="Manage keys-values in file">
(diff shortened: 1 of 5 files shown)


techniques/fileDistribution/manageKeyValueFile/1.0/metadata.xml
```diff
@@ -16,6 +16,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 <!-- This technique lets user define a key-value in a file -->
 <TECHNIQUE name="Manage keys-values in file">
+  <DEPRECATED>This technique version has been superseded by a new version. It will no longer be available in the next stable version of Rudder. Please upgrade to the latest version.</DEPRECATED>
```
amousset: It should not be deprecated in 3.1.

ncharles: good catch !

ncharles:
Commit modified

PR #1200, opened 2017-09-22
Fixes #11405: Add a new syntax for rudder parameters

amousset:
https://www.rudder-project.org/redmine/issues/11405

### DIFF
--- a/techniques/system/common/1.0/rudder-parameters.st
+++ b/techniques/system/common/1.0/rudder-parameters.st
@@ -18,8 +18,42 @@
 
 # This file contain the Global Parameters, to be available within all
 # promises, in the form ${rudder_parameters.parameterName}
+
+# WARNING: This syntax will be deprecated, use the new one defined below ("rudder.parameters[]").
+
 bundle common rudder_parameters {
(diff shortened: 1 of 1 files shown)


amousset:
Commit modified

amousset:
Commit modified

techniques/system/common/1.0/rudder-parameters.st
```diff
@@ -18,8 +18,42 @@
...
+#
+
+# The result key is available in the rudder.<namespace> data variable
+# usage example: ${rudder.parameters[my_global_parameter]}
+#
+bundle agent rudder_global_parameters
+{
+  vars:
+    # The files to read
+    "parameters_file" string => "${this.promise_dirname}/../rudder-parameters.json";
```
ncharles: Why not use "${sys.workdir}/inputs" ?

ncharles: actually, this path is wrong, as it would search for the file in common/rudder-parameters.json

amousset:
Commit modified

PR #1202, opened 2017-10-05
Fixes #10787: Xen domU detection issues with pvops on SLES

armeniaca:


techniques/system/inventory/1.0/fusionAgent.st
```diff
@@ -193,7 +193,7 @@ bundle agent fusionAgent
   vars:
     SuSE.xen_dom0::
       "xen_tools_package" string => "xen-tools";
-    SuSE.xen_domu_pv::
+    SuSE.(xen_domu_pv|xen_domu_hv)::
```
peckpeck: There's a similar line at line 505:     xen.SuSE.xen_domu_pv::
I think it should be fixed too

armeniaca: The issue in redmine targeted the bug against branch 3.1. This PR is targeted against 4.1 by mistake.

I have fixed it on #1203 which is targeted against 3.1. Do you want to upmerge that fix or should I also adjust this PR against 4.1?

armeniaca: well, after checking, I found out it does not merge cleanly from 3.1 to 4.1, so I also modified the PR against 4.1 with the fix. Do you need a separate issue for merging it against 4.1?

peckpeck:
Fixed by #1203

PR #1205, opened 2017-10-12
Fixes #8739: group management technique doesn't ensure optional user is in group

amousset:
https://www.rudder-project.org/redmine/issues/8739


amousset:
OK, merging this PR

PR #1211, opened 2017-10-18
Fixes #11257: Do not send inventory if a forwarder/uploader daemon exists

peckpeck:
https://www.rudder-project.org/redmine/issues/11257

### DIFF
--- a/techniques/system/distributePolicy/1.0/propagatePromises.st
+++ b/techniques/system/distributePolicy/1.0/propagatePromises.st
@@ -189,10 +189,11 @@ bundle agent sendInventoryToCmdb
       "pass3" expression => "pass2";
       "pass2" expression => "pass1";
       "pass1" expression => "any";
+      "disable_upload" expression => fileexists("/opt/rudder/etc/no_inventory_upload");
 
   files:
     # Root server and relay-top send to the endpoint, every other send as usual to their policy server
-    root_server|role_rudder_relay_top::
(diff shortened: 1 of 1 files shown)


techniques/system/distributePolicy/1.0/propagatePromises.st
```diff
@@ -183,6 +183,7 @@ bundle agent sendInventoryToCmdb
     any::
       "incoming_inventory_files" slist => findfiles("${g.rudder_inventories}/incoming/*.ocs", "${g.rudder_inventories}/incoming/*.ocs.gz");
       "updates_inventory_files" slist => findfiles("${g.rudder_inventories}/accepted-nodes-updates/*.ocs", "${g.rudder_inventories}/accepted-nodes-updates/*.ocs.gz");
+      "disable_upload" string => fileexists("/opt/rudder/etc/no_inventory_upload");
```
ncharles: this should be in a classes: section, with the syntax
`"disable_upload" expression => fileexists("/opt/rudder/etc/no_inventory_upload");`

it is used in several bundle, so it should either be in a bundle common, or with a scope namespace

peckpeck: oops indeed for the class
but it is used in a single bundle

peckpeck:
Commit modified

peckpeck:
OK, merging this PR

PR #1215, opened 2017-11-06
Fixes #11687: Create a first implementation of multiversion technique

peckpeck:
https://www.rudder-project.org/redmine/issues/11687

### DIFF
--- a/techniques/fileDistribution/fileTemplate/1.0/fileTemplate.ps1.st
+++ b/techniques/fileDistribution/fileTemplate/1.0/fileTemplate.ps1.st
@@ -1,4 +1,4 @@
-function fileTemplate {
+function fileTemplate_RudderUniqueID {
   [CmdletBinding()]
   param (
       [parameter(Mandatory=$true)]  [string]$reportId,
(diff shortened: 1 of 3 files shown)


peckpeck:
Commit modified

techniques/fileDistribution/fileTemplate/1.0/metadata.xml
```diff
@@ -19,7 +19,7 @@
```
fanf: We also need to specify that the technique support the generation of multiple policy without merge. 
For now, the xml element for that is: 
  <POLICYGENERATION>multi</POLICYGENERATION>

(and a default value of "merge" in place of multi, if the tag is missing, for compatibility)

peckpeck: I'd replace "multi" with "separate"

peckpeck:
Commit modified

peckpeck:
Commit modified

fanf:
Works fine with #1780. Merging it!

techniques/fileDistribution/fileTemplate/1.0/fileTemplate.ps1.st
```diff
@@ -1,4 +1,4 @@
-function fileTemplate {
+function fileTemplate_RudderUniqueID {
```
ncharles: metadata has not been accordingly changed :/

PR #1216, opened 2017-11-06
Fixes #10912: Duration of logrotate for /var/log/rudder/  is too long making /var/log too big

peckpeck:
https://www.rudder-project.org/redmine/issues/10912


peckpeck:
Commit modified

peckpeck:
OK, merging this PR

PR #1224, opened 2017-11-29
Fixes #11803: On update error, the help message is not valid anymore

peckpeck:
https://www.rudder-project.org/redmine/issues/11803


amousset:
Shouldn't this got to 4.1 instead?

peckpeck:
4.1 agent can communicate with 3.1 servers which still have  this issue

amousset:
Once 4.1 promises are in place the agent will never use the old protocol, it may only happen with 3.1 agents with initial promises, which use the 3.1 branch of rudder-techniques

peckpeck:
PR replaced by https://github.com/Normation/rudder-techniques/pull/1225

PR #1232, opened 2017-12-13
Fixes #6564: The technique file copy from shared folder reports file source instead of file destination

ncharles:
https://www.rudder-project.org/redmine/issues/6564


ncharles:
Commit modified

PR #1237, opened 2018-01-04
Fixes #11920: Use rudder agent check in aix crontab

amousset:
https://www.rudder-project.org/redmine/issues/11920

### DIFF
--- a/techniques/system/common/1.0/cron-setup.st
+++ b/techniques/system/common/1.0/cron-setup.st
@@ -58,13 +58,18 @@ bundle agent setup_cronjob
       # Cleanup the pre-4.0 crontab
       "/var/spool/cron/crontabs/root"
         edit_defaults => noempty_backup,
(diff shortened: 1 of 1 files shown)


techniques/system/common/1.0/cron-setup.st
```diff
@@ -58,13 +58,13 @@ bundle agent setup_cronjob
       # Cleanup the pre-4.0 crontab
       "/var/spool/cron/crontabs/root"
         edit_defaults => noempty_backup,
-        edit_line => delete_lines_matching("0,5,10,15,20,25,30,35,40,45,50,55 \* \* \* \* /usr/bin/test -x /opt/rudder/bin/rudder \&\& /opt/rudder/bin/rudder agent check >/dev/null");
+        edit_line => delete_lines_matching("0,5,10,15,20,25,30,35,40,45,50,55 \* \* \* \* /usr/bin/test -x /opt/rudder/bin/check-rudder-agent.*");
```
peckpeck: You should keep all delete_lines so that upgrading from any version works

amousset: I just switched the two line, but I see we also need to add the log file rediretion, and we indeed need tis delete_lines, good catch

amousset:
Commit modified

PR #1238, opened 2018-01-08
Fixes #11929: Remove calls related to expected resports csv files from rudder techniques

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11929

### DIFF
--- a/techniques/system/common/1.0/promises.st
+++ b/techniques/system/common/1.0/promises.st
@@ -89,7 +89,6 @@ bundle common va
         "restart_services", 
 &endif&
 
-        "_clean_current_expected_reports_file",
         "endExecution"
       };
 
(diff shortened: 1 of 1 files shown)


techniques/system/common/1.0/promises.st
```diff
@@ -378,13 +377,8 @@ bundle agent initialize_ncf()
 
   methods:
       "initialize ncf"                  usebundle => initialization;
-      # Force generating expected reports file for Rudder logger
-      # because the first call to the logger may be in audit mode.
-      "create ncf expected reports"     usebundle => _create_current_expected_reports_file;
       !logger_rudder_final_resfile_error::
```
ncharles: you ought to remove this class condition, which is not used anymore

PR #1244, opened 2018-01-26
Fixes #11791: Allow configuring a proxy for HTTP(S) communication bet…

ncharles:
…ween agent and policy server


peckpeck:
We should have a configuration setting for this.

PR #1246, opened 2018-02-08
Fixes #9676: An agent run with initial promises should send its inventory more often

peckpeck:
https://www.rudder-project.org/redmine/issues/9676


peckpeck:
OK, merging this PR

PR #1257, opened 2018-02-20
Fixes #12139: If we a have generic method command_execution with parameter /bin/true, we get an error on reporting for syslog restart on non-aix or solaris sytem

ncharles:
https://www.rudder-project.org/redmine/issues/12139


ncharles:
Commit modified

amousset:
OK, merging this PR

PR #1267, opened 2018-03-20
Fixes #11996: Broken reporting on Rudder 4.3

ncharles:
https://www.rudder-project.org/redmine/issues/11996

### DIFF
--- a/techniques/system/common/1.0/restart-services.st
+++ b/techniques/system/common/1.0/restart-services.st
@@ -81,7 +81,7 @@ bundle agent restart_services
     # We cannot detect which one is used for Rudder
     !windows.!aix.!solaris::
       "run_syslog"      usebundle => disable_reporting;
-      "run_syslog"      usebundle => _method_reporting_context("Make sure syslog service runs", "None");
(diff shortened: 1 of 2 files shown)


techniques/system/common/1.0/restart-services.st
```diff
@@ -81,7 +81,7 @@ bundle agent restart_services
     # We cannot detect which one is used for Rudder
     !windows.!aix.!solaris::
       "run_syslog"      usebundle => disable_reporting;
-      "run_syslog"      usebundle => _method_reporting_context("Make sure syslog service runs", "None");
+      "run_syslog"      usebundle => _method_reporting_context("Log system running", "None");
```
amousset: All other reports about syslog service are on "Log system for reports", why changing it?

ncharles: because i made a mistake :) 
it is a new component though, so we need a name stating that it's running
how about "Log system for reports running"

ncharles:
This was fixed in https://github.com/Normation/rudder-techniques/pull/1260 that prevented report from the rsyslog restart component

PR #1268, opened 2018-03-22
Fixes #12263: Wrong regex in FileTemplate technique

Fdall:
https://www.rudder-project.org/redmine/issues/12263


Fdall:
Commit modified

PR #1272, opened 2018-04-03
Fixes #12348: Correct path to permlist file in File Permissions in 4.3

ncharles:
https://www.rudder-project.org/redmine/issues/12348


ncharles:
Don't merge for now

ncharles:
@amousset if I drop the OUTPATH from the metadata, permlist is correctly generated, but with a .cf - do you think it would be an issue ?

amousset:
What is the current status of this PR?

ncharles:
You may merge it !

PR #1273, opened 2018-04-03
Fixes #12346: Update the techniques tests

Fdall:
https://www.rudder-project.org/redmine/issues/12346


Fdall:
Deleted files which were just copies of v1.0 techniques and so not adapted.
Changed names to make possible test on multi versions techniques.
Some tests are still broken because of their conception (keyValue for exemple which use 2 metadatas and only 1 json, it should use 2 metadatas and 2 json operating on separated files).

ncharles:
Does that means that tests on checkGenericFileContent are working ?

Fdall:
I did not check for tests rightfulness but just for their format.
They are fine on my computer (failing on the initialisation of the 2nd test, meaning the proper execution of the test framework.

However I have a typo error in my commit. Will put a comment and amend.

techniques/fileDistribution/checkGenericFileContent/7.0/tests/simple_content_enforce.metadata
```diff
@@ -1,5 +1,8 @@
-{
-  "directive": "simple_content_enforce.json",
-  "check": "simple_content_enforce.rb", 
+[{
+  "inits": [],
+  "directives": "[simple_content_enforce.json"],
```
Fdall: Typo error

Fdall:
Commit modified

PR #1288, opened 2018-04-25
Fixes #12529: Don't stop agent execution when uuid is missing on non community agent

ncharles:
https://www.rudder-project.org/redmine/issues/12529


techniques/system/common/1.0/update.st
```diff
@@ -387,11 +387,13 @@ bundle agent update_action
...
-    "should_not_continue" expression => "wrong_uuid|no_uuid";
-    "no_uuid" not => fileexists("${g.uuid_file}");
+      # Abort if no uuid is defined
+      "should_not_continue" expression => "wrong_uuid|no_uuid";
+    community_edition::
+      "no_uuid" not => fileexists("${g.uuid_file}");
 &if(!INITIAL)&
-    "wrong_uuid" not => strcmp("&UUID&", "${g.uuid}");
+    any::
+      "wrong_uuid" not => strcmp("&UUID&", "${g.uuid}");
```
amousset: how can `g.uuid` be defined if `uuid_file` does not exist?

amousset: yes but doesn't g.uuid come from uuid_file?

ncharles: yes

ncharles: ha, you want to check it only if file is there ?

ncharles:
Commit modified

PR #1297, opened 2018-05-18
Fixes #12262: jobScheduler does not work when the first non-loopback interface has no ip address

amousset:
https://www.rudder-project.org/redmine/issues/12262


amousset:
OK, merging this PR

PR #1302, opened 2018-05-23
Fixes #12687: Add an entry for the slackware package module in the packageManagement technique metadata

VictorQrt:
https://www.rudder-project.org/redmine/issues/12687


VinceMacBuche:
omg, why was it merged ???

PR #1309, opened 2018-05-30
Fixes #11683: Process management: stopped sends multiple reports.

Fdall:
https://www.rudder-project.org/redmine/issues/11683


ncharles:
Hum, the code is misleading - start/stop/restart are indeed base on process name

Fdall:
Commit modified

PR #1321, opened 2018-07-03
Fixes #6343: When a technique reports several time the exact same text, there is only one report generated

ncharles:
https://www.rudder-project.org/redmine/issues/6343


amousset:
This will need to be done in 4.3 too.

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1342, opened 2018-09-17
Fixes #13491: deprecate technique nfsServer in Rudder 5.0

ncharles:
https://www.rudder-project.org/redmine/issues/13491


techniques/systemSettings/networking/nfsServer/3.0/metadata.xml
```diff
@@ -15,6 +15,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 
 <TECHNIQUE name="NFS server">
+  <DEPRECATED>This technique version has been superseded by a new version. It will no longer be available in the next stable version of Rudder. Please upgrade to the latest version.</DEPRECATED>
```
amousset: The message should be different, this technique will be totally removed.

ncharles:
Commit modified

PR #1344, opened 2018-10-11
Fixes #13646: Add a bigger warning about full path in file fomr rudder technique

peckpeck:
https://www.rudder-project.org/redmine/issues/13646


peckpeck:
OK, merging this PR

PR #1345, opened 2018-10-11
Fixes #13647: Add a bigger warning about full path in file from rudder technique

peckpeck:
https://www.rudder-project.org/redmine/issues/13647


peckpeck:
OK, merging this PR

PR #1360, opened 2018-11-12
Fixes #13290: Broken reporting when copying a directory from the Rudder server without specifying the trailing / in the destination

amousset:
https://issues.rudder.io/issues/13290

### DIFF
--- a/techniques/fileDistribution/copyGitFile/2.2/copyFileFromSharedFolder.st
+++ b/techniques/fileDistribution/copyGitFile/2.2/copyFileFromSharedFolder.st
@@ -52,6 +52,10 @@ bundle agent download_from_shared_folder
       &COPYFILE_DESTINATION:{destination |"copyfile[&i&][destination]" string => "&destination&";
 }&
 
+      # Raw version for reporting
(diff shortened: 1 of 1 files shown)


techniques/fileDistribution/copyGitFile/2.2/copyFileFromSharedFolder.st
```diff
@@ -52,6 +52,10 @@ bundle agent download_from_shared_folder
       &COPYFILE_DESTINATION:{destination |"copyfile[&i&][destination]" string => "&destination&";
 }&
 
+      # Raw version for reporting
+      &COPYFILE_DESTINATION:{destination |"copyfile[&i&][raw_destination]" string => "&destination&";
+}&
```
ncharles: shouldn't you change l66 to "canon_dirname_${index}" string => canonify(dirname("${copyfile[${index}][raw_destination]}")); ?

ncharles: actually, no, that's correct. Thank you

PR #1365, opened 2018-11-13
Fixes #13819: Remove default server from system technique

peckpeck:
https://issues.rudder.io/issues/13819

### DIFF
--- a/techniques/system/common/1.0/cf-serverd.st
+++ b/techniques/system/common/1.0/cf-serverd.st
@@ -73,7 +73,7 @@ bundle server access_rules
         comment            => "Grant ${query_types} reporting query for the hub on the policy server",
         resource_type      => "query",
         report_data_select => rudder_data_select_policy_hub,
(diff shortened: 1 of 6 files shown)


techniques/system/common/1.0/cf-serverd.st
```diff
@@ -73,7 +73,7 @@ bundle server access_rules
         comment            => "Grant ${query_types} reporting query for the hub on the policy server",
         resource_type      => "query",
         report_data_select => rudder_data_select_policy_hub,
-        admit_ips          => { host2ip("${def.policy_server}"), @{sys.ip_addresses} }; # an enterprise policy server needs to be able to contact itself
+        admit_ips          => { host2ip("${server_info.policy_server}"), @{sys.ip_addresses} }; # an enterprise policy server needs to be able to contact itself
```
ncharles: policy file cf-serverd is loaded before file update.cf
are we sure that the agent knows how to resolve these variable in the right order ?

peckpeck: I'm pretty sure it works since this variable is already used below, line 102

PR #1371, opened 2018-11-19
Fixes #13859: Load all ncf files in one go (or two)

ncharles:
https://issues.rudder.io/issues/13859


ncharles:
PR replaced by https://github.com/Normation/rudder-techniques/pull/1372

PR #1374, opened 2018-11-21
Fixes #13887: Option descrition for "dowload for shared files" is mileading and mixed use of folder/directory

fanf:
https://issues.rudder.io/issues/13887


fanf:
OK, merging this PR

PR #1376, opened 2018-12-03
Fixes #13932: copyGitFile technique in very old version does not work anymore since a common variable has been removed

Fdall:
https://issues.rudder.io/issues/13932


techniques/system/common/1.0/update.st
```diff
@@ -108,6 +108,8 @@ bundle common server_info
       "policy_server" string => ifelse(fileexists("${policy_server_file}"),
                                   readfile("${policy_server_file}", 2048),
                                   "");
+      "cfserved"      string => "${policy_server}";
```
peckpeck: a comment on the reason of this variable and that it should be kept would have been welcome

PR #1381, opened 2018-12-14
Fixes #13975: Add compatibility sles classes in failsafe

amousset:
https://issues.rudder.io/issues/13975


ncharles:
you hsould not change path of apache-vhost.tpl

techniques/system/common/1.0/failsafe.st
```diff
@@ -126,6 +126,12 @@ bundle common g
 
 # definition of the node roles
   &NODEROLE&
+
+  classes:
+      # OS classes for compatibility
```
peckpeck: shouldn't there be a scope => namespace ?

amousset: Nope, it is a bundle common where classes are global by default.

amousset:
ping @peckpeck

PR #1394, opened 2018-12-24
Fixes #14038: Separated policy generation does not work correctly with Rudder DSC agent

ncharles:
https://issues.rudder.io/issues/14038


ncharles:
must be merged after https://github.com/Normation/rudder/pull/2103

PR #1399, opened 2019-01-09
Fixes #14124: Remove rudder-lib from techniques

amousset:
https://issues.rudder.io/issues/14124


amousset:
OK, squash merging this PR

PR #1411, opened 2019-02-12
Fixes #14296: System technique try to edit pg_hba file only on SUSE, but packaging takes care of this part

ncharles:
https://issues.rudder.io/issues/14296


techniques/system/server-roles/1.0/postgres-check.st
```diff
@@ -1,69 +0,0 @@
...
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-#####################################################################################
-
-bundle agent root_postgres_check
-{
-
-  vars:
-
-    SuSE::
```
amousset: Wasn't this an exception for SLES?

ncharles: this is done here, on all rpm systems
https://github.com/Normation/rudder-packages/blob/branches/rudder/4.1/rudder-reports/SPECS/rudder-reports.spec#L163

amousset: but do you know why there was an exception for SLES?

ncharles: it's from the old time https://issues.rudder.io/issues/2207

ncharles:
retargeting to 5.0

ncharles:
PR replaced by https://github.com/Normation/rudder-techniques/pull/1424

PR #1418, opened 2019-02-20
Fixes #14353: Remove unencrypted body files

amousset:
https://issues.rudder.io/issues/14353

### DIFF
--- a/techniques/system/common/1.0/update.cf
+++ b/techniques/system/common/1.0/update.cf
@@ -16,7 +16,6 @@
 #
 #####################################################################################
 
-
 #
 # Define what an agent has to do when it updates
 # its promises
@@ -25,7 +24,6 @@
 # Use for sensitive files
 body copy_from remote(server, path)
 {
-
         servers    => {
           "${server}"
         };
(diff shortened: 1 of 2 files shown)


techniques/system/common/1.0/update.cf
```diff
@@ -25,11 +24,9 @@
 # Use for sensitive files
 body copy_from remote(server, path)
 {
-
         servers    => {
           "${server}"
         };
-        encrypt    => "true";
```
peckpeck: Is that the default ?

amousset:
Commit modified

techniques/system/common/1.0/update.cf
amousset: Good spot it is not actually. This has practically no impact as "When used with protocol_version 2 or greater this attribute is a noop as the entire session is encrypted." but who knows, it's probaly better to keep it anyway.

techniques/system/common/1.0/update.cf
```diff
@@ -74,30 +51,9 @@ body copy_from copy_digest_without_perms(from)
         portnumber => "${system_common.community_port}";
 }
 
-# Use for non-sensitive policy files
-body copy_from remote_unsecured_without_perms(server, path)
```
ncharles: this is used in techniques/system/distributePolicy/1.0/propagatePromises.st - you should remove it there also

amousset:
Commit modified

amousset:
OK, squash merging this PR

PR #1426, opened 2019-04-03
Fixes #14581: Errors at Rudder 5 installation about missing policy_server.dat

ncharles:
https://issues.rudder.io/issues/14581

### DIFF
--- a/techniques/system/common/1.0/update.st
+++ b/techniques/system/common/1.0/update.st
@@ -104,14 +104,23 @@ bundle common server_info
         string  => translatepath("${sys.workdir}/policy_server.dat"),
         comment => "Path to file containing address to policy server";
 
+    pass2.policy_server_file_exists::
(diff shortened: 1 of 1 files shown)


techniques/system/common/1.0/update.st
```diff
@@ -104,14 +104,23 @@ bundle common server_info
...
+    pass2::
       "cfserved"      string => "${policy_server}";
 
   classes:
-      "should_disable_server" expression => fileexists("${g.rudder_disable_server_file}");
+      "should_disable_server"     expression => fileexists("${g.rudder_disable_server_file}");
+
+      "policy_server_file_exists" expression => fileexists("${policy_server_file}");
+
+      "pass2" expression => "pass1";
```
ncharles: it should not have pass2 here !!!! these are GLOBAL classes

ncharles: and it breaks 14592 because having to define variable in bundle common with two passes fails

PR #1430, opened 2019-04-23
Fixes #14732: sshKeyDistribution creates home directory with perms root:root when it does not exists yet 

ncharles:
https://issues.rudder.io/issues/14732


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1433, opened 2019-04-23
Fixes #14750: If postgresql is down, agent tried to fix password before checking that its up an running

ncharles:
https://issues.rudder.io/issues/14750


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1439, opened 2019-05-09
Fixes #14845: Remoge key generation from system techniques

peckpeck:
https://issues.rudder.io/issues/14845


techniques/system/common/1.0/failsafe.st
```diff
@@ -93,20 +93,12 @@ bundle agent init_files
     community_edition::
       "cfengine_install_path" string => "${g.rudder_base}";
 
-  classes:
-      "missing_key" not => fileexists("${sys.workdir}/ppkeys/localhost.priv");
```
amousset: I had no idea we had this here! (and it generates smaller keys, as it was not modified like the other calls)

PR #1450, opened 2019-06-05
Fixes #15028: Improve 'rsyslog' configuration for insanely high load 

ncharles:
https://issues.rudder.io/issues/15028


amousset:
Is the effect visible?

ncharles:
yes, i gain about 10-20% CPU usage for 1800reports/s non-stop

PR #1455, opened 2019-07-05
Fixes #15163: User Management v9 doesn't report when we request a change of user home directory and it fails to change its home

ncharles:
https://issues.rudder.io/issues/15163


ncharles:
Commit modified

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1470, opened 2019-07-23
Fixes #15282: Rudder commands complain because of a missing rudder.json

peckpeck:
https://issues.rudder.io/issues/15282


amousset:
why closing it?

PR #1485, opened 2019-08-01
Fixes #15380: Add the User techniques category to the technique packaging

peckpeck:
https://issues.rudder.io/issues/15380

### DIFF
--- a/techniques/ncf_technique/category.xml
+++ b/techniques/ncf_technique/category.xml
@@ -1,0 +1,6 @@
+<xml>           
+  <name>User Techniques</name>
+  <description>
+    Techniques created using the Technique editor.
+  </description>
+</xml>


techniques/ncf_technique/category.xml
```diff
@@ -0,0 +1,6 @@
+<xml>           
+  <name>User Techniques</name>
```
amousset: This will break if people already have it with a different name IIRC

peckpeck: We do not push categories to existing configuration-repository directory

amousset: won't it be updated by techique-upgrade mechanism?

peckpeck: Then the name will be updated like for other categories.

amousset: the name is not updated when changed in the category file, but a new category is created. This is an old known bug.

peckpeck: %µ$£ù#*

peckpeck: Now it won't anymore

amousset: \o/

amousset:
Needs to be retargeted in 5.1 (and te remove this category creation from the postinstall)

peckpeck:
PR replaced by https://github.com/Normation/rudder-techniques/pull/1503

PR #1490, opened 2019-08-07
Fixes #15424: ncf is not updated anymore

ncharles:
https://issues.rudder.io/issues/15424


bootstrap-promises/failsafe.cf
```diff
@@ -123,6 +125,12 @@ body depth_search recurse(d)
     depth => "${d}";
 }
 
+body depth_search recurse_ignore(d,list)
```
amousset: this should go into ncf

PR #1495, opened 2019-08-13
Fixes #15458: Adapt techniques to stop agent if it's not in audit and the audit is enforced

ncharles:
https://issues.rudder.io/issues/15458

### DIFF
--- a/techniques/system/common/1.0/promises.st
+++ b/techniques/system/common/1.0/promises.st
@@ -38,6 +38,8 @@ bundle common rudder_roles
       "rudder_reporting_https" expression => strcmp("&REPORTING_PROTOCOL&", "HTTPS");
       "rsyslog_disabled"       expression => strcmp("&SYSLOG_PROTOCOL_DISABLED&", "true");
 
+      "rudder_global_audit_mode"      not => strcmp("&RUDDER_COMPLIANCE_MODE&", "enforce");
+
       # full compliance is the default mode
       "full_compliance" not => "changes_only|reports_disabled";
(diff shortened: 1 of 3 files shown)


techniques/system/common/1.0/promises.st
```diff
@@ -235,9 +238,34 @@ bundle agent check_disable_agent
...
+# All promises in this bundle MUST have action=>immediate,
+# to avoid locking issues in case of several close runs.
+##########################################################
+bundle agent check_rudder_force_audit_agent
+{
+  classes:
+      "should_be_in_audit" expression => fileexists("${g.rudder_force_audit_agent}");
+
+      # Only define this class when we're ready to die - this is a special class name in "abortclasses"
+      "should_not_continue"
```
amousset: could you use the new abort mechanism instead?

amousset: Like in ```bundle agent rudder_check_agent_version   
{
  methods:  
    cfengine_3_1|cfengine_3_2|cfengine_3_3|cfengine_3_4|cfengine_3_5|cfengine_3_6|cfengine_3_7|cfengine_3_8|cfengine_3_9::
      "any" usebundle => _abort("unsupported_agent", "This agent is not compatible with its Rudder server, please upgrade");
}
```

amousset: (it send a control report to the webapp before aborting, which may be useful in the future)

ncharles:
Commit modified

techniques/system/common/1.0/promises.st
ncharles: i had no idea it even existed

PR #1497, opened 2019-08-19
Fixes #15496: missing report about  	Send inventories to Rudder server

ncharles:
https://issues.rudder.io/issues/15496


ncharles:
ha, report is still not in web interface

ncharles:
ha, just needed a bit of time for the web interface to get the report

PR #1498, opened 2019-08-19
Fixes #15513: Make certificate verification in HTTP calls configurable

amousset:
https://issues.rudder.io/issues/15513


ncharles:
looks good to me, but we need the scala code to be able to merge it, or it will break rudder

amousset:
PR updated with a new commit

amousset:
OK, squash merging this PR

PR #1500, opened 2019-08-20
Fixes #15523: Relayd configuration is world-readable

amousset:
https://issues.rudder.io/issues/15523


ncharles:
the Makefile should also be updated:
        install -m 644 relayd/tools/config/main.conf $(DESTDIR)/opt/rudder/etc/relayd/main.conf
        install -m 644 relayd/tools/config/logging.conf $(DESTDIR)/opt/rudder/etc/relayd/logging.conf

it should also be 640

PR #1517, opened 2019-09-30
Fixes #15806: Agent should not try to set permission of certificate if it is a symbolic link

ncharles:
https://issues.rudder.io/issues/15806


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1518, opened 2019-10-01
Fixes #15818: Bootstrap policies should listen on 5309

amousset:
https://issues.rudder.io/issues/15818


amousset:
OK, squash merging this PR

PR #1421, opened 2019-02-28
Fixes #14369: timezone technique does not support systemd

Fdall:
https://issues.rudder.io/issues/14369

PR #1526, opened 2019-10-03
Fixes #14369: timezone technique does not support systemd

Fdall:
https://issues.rudder.io/issues/14369

Replacing previous PR: https://github.com/Normation/rudder-techniques/pull/1421


fanf:
There was internal discussion on that one, and it's not ready for prime time. @ncharles and @Fdall need to decide what is the correct next steps on it.

fanf:
This is too hard to maintain given difference in ntp/crony/systemd 
We will make one:
- for timezone
- for time synchronization (ntp/crony + systemd)

They are still techniques, not generic method.

PR #1539, opened 2019-10-21
Fixes #16002: When switching to HTTPS reporting, the compliance of root server is broken because it receives both https and syslog

ncharles:
https://issues.rudder.io/issues/16002


techniques/system/distributePolicy/1.0/rudder-rsyslog-relay.st
```diff
@@ -30,7 +30,7 @@ $ActionQueueSaveOnShutdown on
 # If report protocol is HTTPS, we drop the local rsyslog message
 {{#classes.rudder_reporting_https}}
 if $fromhost-ip == "127.0.0.1" then {
-    :programname, isequal, "rudder" ~
```
amousset: We should maybe support both for compatibility

ncharles: make sense

ncharles:
Commit modified

PR #1546, opened 2019-10-22
Fixes #16028: Initial policies on root server broken with missing postgres-check.cf file

amousset:
https://issues.rudder.io/issues/16028


amousset:
OK, merging this PR

PR #1555, opened 2019-11-13
Fixes #15032: Add capabilities as system classes

amousset:
https://issues.rudder.io/issues/15032


techniques/system/common/1.0/promises.st
```diff
@@ -61,12 +61,20 @@ bundle agent rudder_init {
...
       "Load properties"                usebundle => properties;
 &endif&
 
 }
 
+bundle agent load_agent_capabilities {
+  vars:
+      "capabilities" slist => readstringlist("${g.rudder_capabilities_file}", "", "\n", inf, inf);
+  classes:
+      "agent_capability_${capabilities}";
```
ncharles: these classes are local, what is their use ?

PR #1558, opened 2019-11-13
Fixes #16178: Agent is not correctly aborted when repaired is happening in audit mode

amousset:
https://issues.rudder.io/issues/16178


amousset:
OK, merging this PR

PR #1559, opened 2019-11-13
Fixes #11086: File content directive - Audit mode is not correctly supported

ncharles:
https://issues.rudder.io/issues/11086


techniques/fileDistribution/checkGenericFileContent/8.0/checkGenericFileContent.st
```diff
@@ -87,6 +87,15 @@ bundle agent check_generic_file_content_RudderUniqueID {
     "generic_file_content_creation[&i&]" string => "true";
 }&
 
+    pass2.dry_run::
+      # we need to sanitize the repaired classes of replace_pattern in audit mode
```
amousset: could you add a comment with a link to the issue, and state that this should be removed once broken agents are replaced?

ncharles:
Commit modified

PR #1573, opened 2020-01-06
Fixes #12990: Add Amazon Linux support to ncf and techniques

amousset:
https://issues.rudder.io/issues/12990


amousset:
Commit modified

PR #1576, opened 2020-01-09
Fixes #16518: Missing report on \"Synchronize files\" on simple relay…

amousset:
…s when shared-files are empty


amousset:
OK, merging this PR

PR #1578, opened 2020-01-14
Fixes #16549: Shared files are broken in relayd

amousset:
https://issues.rudder.io/issues/16549


techniques/system/server-roles/1.0/relayd.conf.tpl
```diff
@@ -73,7 +73,7 @@ command = "/opt/rudder/bin/rudder"
 use_sudo = true
 
 [shared_files]
-path = "{{{vars.system_common.sharedfiles_folder}}}"
```
amousset: this one actually had the same value as `vars.g.shared_files`. We are paying our poor directory naming I guess.

PR #1586, opened 2020-03-03
Fixes #16849: When using twice directive packageManagement 1.2, once to ensure presence of a package, and second one to upgrade the package, reporting and posthook is wrong on the second one

ncharles:
https://issues.rudder.io/issues/16849


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1595, opened 2020-05-18
Fixes #17430: reports and inventory cannot be send to relay server

VinceMacBuche:
https://issues.rudder.io/issues/17430

### DIFF
--- a/techniques/system/server-roles/1.0/password-check.cf
+++ b/techniques/system/server-roles/1.0/password-check.cf
@@ -219,7 +219,7 @@ bundle agent root_password_check_dav
 
   classes:
 
-      "dav_cant_connect" not => returnszero("${g.rudder_curl} --tlsv1.2 --proxy '' ${g.rudder_verify_certs_option} --silent --fail --output /dev/null --user ${g.davuser}:${p.dav_password[2]} --upload-file /opt/rudder/etc/uuid.hive https://localhost/inventory-updates/uuid.hive","noshell");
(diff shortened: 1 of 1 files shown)


techniques/system/server-roles/1.0/password-check.cf
```diff
@@ -219,7 +219,7 @@ bundle agent root_password_check_dav
 
   classes:
 
-      "dav_cant_connect" not => returnszero("${g.rudder_curl} --tlsv1.2 --proxy '' ${g.rudder_verify_certs_option} --silent --fail --output /dev/null --user ${g.davuser}:${p.dav_password[2]} --upload-file /opt/rudder/etc/uuid.hive https://localhost/inventory-updates/uuid.hive","noshell");
+      "dav_cant_connect" not => returnszero("${g.rudder_curl} --tlsv1.2 --proxy '' ${g.rudder_verify_certs_option} --silent --fail --output /dev/null --user ${g.davuser}:${g.davpw} --upload-file /opt/rudder/etc/uuid.hive https://localhost/inventory-updates/uuid.hive","noshell");
```
amousset, line 222: this won't work immediately on root, we can maybe do a special case for it.

VinceMacBuche: this work for root, same password as generated id ok, we only use value in rudder-passwords.conf in this file, to generate rudder properties

VinceMacBuche: Juste we don't set the new password now before it can be propagated to all nodes (needs a new generation to do so), but it's better to have hte same password everywhere

PR #1598, opened 2020-05-23
Fixes #17478: properties loading needs to be done in a common bundle

amousset:
https://issues.rudder.io/issues/17478


techniques/system/common/1.0/properties.cf
```diff
@@ -63,7 +63,7 @@ bundle agent properties
 }
 
 # Read the inventories variables from Rudder
-bundle agent rudder_node_inventory {
+bundle common rudder_node_inventory {
```
amousset, line 66: not stricly necessary but better be consistent

PR #1605, opened 2020-06-12
Fixes #17722: Initial promises warning during agent setup

peckpeck:
https://issues.rudder.io/issues/17722


ncharles:
isn't this file common for all nodes, and so for root server during its install?

peckpeck:
It is, but even the root server doesn't need to serve files during its initial setup

PR #1607, opened 2020-06-17
Fixes #17765: error when installing rudder server root on debian 9 - 6.1

ncharles:
https://issues.rudder.io/issues/17765

### DIFF
--- a/variables.json
+++ b/variables.json
@@ -15,7 +15,7 @@
   "MANAGED_NODES_NAME": [ "localhost" ],
   "MODIFIED_FILES_TTL": "30",
   "NODEROLE": "# This node doesn't have any specific role",
-  "NOVA": true,
+  "NOVA": false,
   "POLICYSERVER_ADMIN": "root",
   "RUDDER_DIRECTIVES_INPUTS": "",
   "RUDDER_DIRECTIVES_SEQUENCE": "",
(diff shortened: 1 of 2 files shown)


variables.json
```diff
@@ -36,5 +36,5 @@
   "STORE_ALL_CENTRALIZED_LOGS_IN_FILE": true,
   "SYSLOGPORT": "${install_rsyslogd.rsyslog_port[2]}",
   "TRACKINGKEY": "root-DP@@root-distributePolicy@@00",
-  "RUDDER_INVENTORY_VARS": "{}"
+  "RUDDER_INVENTORY_VARS": "{ "initial-policies":"true" }"
```
ncharles, line 39: i didn't want to disable the rudder inventory vars in initial policies,, as they may turn out to be useful in the future

peckpeck: what about the "INITIAL" variable that already exists in this file ?

peckpeck: I don't understand this line, what in the usage of this inventory variable ?

ncharles: the only goal here is to avoir having an error at installation (read the related ticket).
i'm pretty sure we'll want to have content in there in the future, so i didn't want to remove the reading of this content in initial policies

PR #1642, opened 2020-10-30
Fixes #18449: ssh key distribution techniques doesn't accept - in user login

ncharles:
https://issues.rudder.io/issues/18449


erenfro:
I can confirm after manually updating my local environment, this functions as expected, at least at the end result.

amousset (approved):

PR #1645, opened 2020-11-25
Fixes #18695: Create a technique to manage yum repository

ncharles:
https://issues.rudder.io/issues/18695

### DIFF
--- a/techniques/applications/yumPackageManagerRepositories/1.0/yum-repositories-management.st
+++ b/techniques/applications/yumPackageManagerRepositories/1.0/yum-repositories-management.st
@@ -1,0 +1,81 @@
+bundle agent yum_repositories_management_&RudderUniqueID& {
+
+  vars:
+      &YUM_REPO_ID:{repo_id |"repo_id[&i&]" string => "&repo_id&";
+}&
+      &YUM_REPO_NAME:{repo_name |"repo_name[&i&]" string => "&repo_name&";
+}&
+      &YUM_REPO_ACTION:{repo_action |"repo_action[&i&]" string => "&repo_action&";
+}&
+
+      &YUM_REPO_URL:{repo_url |"repo_url[&i&]" string => "&repo_url&";
+}&
+      &YUM_REPO_ENABLED:{repo_enabled |"repo_enabled[&i&]" string => "&repo_enabled&";
+}&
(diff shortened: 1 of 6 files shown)


ncharles:
Yum repos have several option for url:
mirrorlist
baseurl

Should we let the user choose which field to set ?

ncharles:
Commit modified

techniques/applications/yumPackageManagerRepositories/1.0/yum-repositories-management.st
```diff
@@ -0,0 +1,81 @@
...
+
+      # report
+      # Using old_class_prefix to avoid having to copy all classes. Drawback is
+      # if there are several repos managed in the same file, there will be
+      # classes colision
+      "report for modification"    usebundle => rudder_common_reports_generic_index("Package sources (Yum)", "file_key_value_present_in_ini_section_${canonified_filename[${index}]}", "${trackingkey[${index}]}", "Repositories", "${repo_id[${index}]}", "Setting repository ${repo_name[${index}]} with id ${repo_id[${index}]} ", "${index}"),
+                                          if => "!delete_repo_${index}";
+                                          
+      # Deletion of repo
+      "delete repo ${index}"      usebundle => file_absent("${filename[${index}]}"),
```
ncharles, line 77: this is overzealous as it may delete several repos if they are in the same file

ncharles: i could make a file promise, that would select the region, and delete all lines matching *, plus deleting the region.
but it may leave nearly empty files

So another idea would be to count the section in a file, with a grep -rc '^\[.*\]$' /etc/yum.repos.d/
if count is 1, then delete the file, else clean it

this would probably be another bundle

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

ncharles:
still not totally functionning, as the deletion doesn't report anything nor delete the file

ncharles:
it's because the section is not present, so nothing to remove, so nothing happens

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

ncharles:
ready for review

ncharles:
PR updated with a new commit

peckpeck:
this one too, is it still valid ?

PR #1654, opened 2021-03-30
Fixes #19089: Add a rudder webapp system technique

Fdall:
https://issues.rudder.io/issues/19089

### DIFF
--- a/techniques/system/rudderWebapp/1.0/integrity-check.cf
+++ b/techniques/system/rudderWebapp/1.0/integrity-check.cf
@@ -1,0 +1,64 @@
+####################################################################################
+# Copyright 2021 Normation SAS
+####################################################################################
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, Version 3.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+bundle agent system_root_integrity_check(trackingkey, technique_name)
+{
+  vars:
+      "key_destination" string => "${sys.workdir}/ppkeys/root-${sys.key_digest}.pub";
+      "component"       string => "Check configuration-repository";
+      "component_key1"  string => "Configuration-repository exists";
+      "component_key2"  string => "Check configuration-repository GIT lock";
+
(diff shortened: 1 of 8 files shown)


Fdall:
PR updated with a new commit

Fdall:
PR updated with a new commit

amousset:
Missing license in new .cf files

Fdall:
PR updated with a new commit

techniques/system/rudderWebapp/1.0/integrity-check.cf
```diff
@@ -0,0 +1,64 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+bundle agent system_root_integrity_check(trackingkey, technique_name)
```
ncharles, line 18: you could fetch the values from bundle system_rudder_webapp, wouldn't it be better ?

techniques/system/rudderWebapp/1.0/integrity-check.cf
```diff
@@ -0,0 +1,64 @@
...
+      "pass2" expression => "pass1";
+      "pass1" expression => "any";
+
+  files:
+      "${g.rudder_configuration_repository}"
+        create => "true",
+        action => WarnOnly,
+        classes => if_else("rudder_integrity_ok", "rudder_integrity_failed");
+
+      "${g.rudder_configuration_repository}/.git/index.lock"
```
ncharles, line 38: What happens if we have a slow git command running ?

ncharles (commented):
Looks good to me - i'm simply unsure about passing the technique name and the tracking key

techniques/system/rudderWebapp/1.0/integrity-check.cf
Fdall: I believe that we should avoid when possible to rely on data taken from other resources.

techniques/system/rudderWebapp/1.0/integrity-check.cf
Fdall: I believe that we should avoid when possible to rely on data taken from other resources.

techniques/system/rudderWebapp/1.0/integrity-check.cf
Fdall: Wrong comment, I have no idea, I just took the existing techniques and port them.

Fdall:
PR updated with a new commit

ncharles (approved):

Fdall:
OK, squash merging this PR

PR #1658, opened 2021-04-01
Fixes #19113: The logrotate conf file still uses legacy service reloading

Fdall:
https://issues.rudder.io/issues/19113

### DIFF
--- a/techniques/system/server-roles/1.0/rudder-logrotate
+++ b/techniques/system/server-roles/1.0/rudder-logrotate
@@ -15,7 +15,7 @@
         delaycompress
         sharedscripts
         postrotate
-          ${root_logrotate_check.service_restart_utility} ${root_logrotate_check.apache_service_name} reload > /dev/null || true
+          ${paths.path[systemctl]} reload ${root_logrotate_check.apache_service_name} > /dev/null || true
         endscript
 }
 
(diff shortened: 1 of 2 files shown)


techniques/system/server-roles/1.0/rudder-logrotate
```diff
@@ -59,7 +59,7 @@
         create ${root_logrotate_check.syslog_file_mode} ${root_logrotate_check.syslog_user} ${root_logrotate_check.syslog_group}
         delaycompress
         postrotate
-          ${root_logrotate_check.service_restart_utility} ${root_logrotate_check.syslog_service_name} ${root_logrotate_check.syslog_rotation_method} > /dev/null || true
+          ${paths.path[systemctl]} ${root_logrotate_check.syslog_rotation_method} ${root_logrotate_check.syslog_service_name} > /dev/null || true
```
amousset, line 62: does `systemctl rotate` exist?

ncharles: also on sles12sp1, is there a systemctl ?

Fdall: You are right, this won't work on debian based system where we are curently calling the init rotate command which send a HUP signal to syslog.
I will close this PR as it is not as straightforward as expected.

Fdall:
Closing this pr since it may break things and only bring non yet mandatory changes.

techniques/system/server-roles/1.0/rudder-logrotate
amousset: SIGHUP is generally "systemctl reload"

PR #1664, opened 2021-05-14
Fixes #19264: Broken inventory upload from relays

amousset:
https://issues.rudder.io/issues/19264


ncharles (commented):
I'm unsure about this really big change.  It's been considered a failsafe macanism for relayd, and i'm not convinced removing it in a patch version is a good idea
I'd rather have a file_select for forwarding selecting files older than x minutes, x being something big, like 30 or 60, so that we are sure the relay won't try to catch them

amousset:
@ncharles  Are you ok to retarget this to 7.0?

ncharles:
Yes !

PR #1673, opened 2021-06-07
Fixes #19403: User technique should show unix option by default

peckpeck:
https://issues.rudder.io/issues/19403


amousset (approved):

amousset:
Makes sense as the Windows version is probably never used anyway

PR #1682, opened 2021-07-19
Fixes #19541: Remove automatic roles management

amousset:
https://issues.rudder.io/issues/19541

### DIFF
--- a/techniques/system/common/1.0/failsafe.st
+++ b/techniques/system/common/1.0/failsafe.st
@@ -64,22 +64,13 @@ bundle common g
       # We would like to use date's "--rfc-3339=second" option here, but it is not available on older OSes (RHEL 3/4, AIX 5...)
       "execRun"                  string => execresult("/bin/date -u \"+%Y-%m-%d %T+00:00\"", "noshell");
 
-# definition of the node roles
-  &NODEROLE&
-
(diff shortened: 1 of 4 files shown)


techniques/system/common/1.0/failsafe.st
```diff
@@ -64,20 +64,13 @@ bundle common g
       # We would like to use date's "--rfc-3339=second" option here, but it is not available on older OSes (RHEL 3/4, AIX 5...)
       "execRun"                  string => execresult("/bin/date -u \"+%Y-%m-%d %T+00:00\"", "noshell");
 
-# definition of the node roles
-  &NODEROLE&
```
ncharles, line 68: this will be a problem on relay: it's the NODEROLE that says that it's a policy server

ncharles (changes requested):
this ought to be ok, but failsafe has no way to know if the node is a policy server or not

techniques/system/common/1.0/failsafe.st
amousset: this information does not seem to be actually used in failsafe

amousset:
PR updated with a new commit

ncharles (approved):

PR #1690, opened 2021-07-26
Fixes #19560: Rename the AUTHORIZED_NETWORKS parameter used in the common system techniques.

Fdall:
https://issues.rudder.io/issues/19560


techniques/system/common/1.0/cf-serverd.st
```diff
@@ -92,9 +92,9 @@ bundle common def
     policy_server::
       "acl" slist => {
         "127.0.0.0/8" , "::1", @{sys.ip_addresses},
-&if(AUTHORIZED_NETWORKS)&
+&if(ALLOWEDNETWORKS)&
```
ncharles, line 95:
```suggestion
&if(ALLOWED_NETWORKS)&
```

ncharles (changes requested):
it is not ALLOWEDNETWORKS but ALLOWED_NETWORKS

techniques/system/common/1.0/metadata.xml
```diff
@@ -76,7 +76,6 @@
   </FILES>
 
   <SYSTEMVARS>
-    <NAME>AUTHORIZED_NETWORKS</NAME>
```
fanf, line 79: it's still used in common technique as a system var, you need to call it `ALLOWED_NETWORKS` here

Fdall:
Commit modified

fanf (approved):

fanf (approved):

PR #1708, opened 2021-08-02
Fixes #19655: Invalid /root/.pgpass file in rudder 7

ncharles:
https://issues.rudder.io/issues/19655


techniques/system/rudder-service-postgresql/1.0/password-check.st
```diff
@@ -24,10 +24,10 @@ bundle agent system_rudder_postgresql(trackingkey, technique_name, db_pass) {
 
     any::
     # Array used to force content of the .pgpass file
-      "pgpass[1]" string => "${rudder_postgresql.host}";
+      #"pgpass[1]" string => "${rudder_postgresql.host}";
```
amousset, line 27: I'm not sure it really makes sense to support a different db name if we don't support a different host

ncharles: i hope to add support of a different host - see https://issues.rudder.io/issues/19659

amousset (approved):

PR #1721, opened 2021-09-07
Fixes #19852: Jetty is restarted 3 times

ncharles:
https://issues.rudder.io/issues/19852

### DIFF
--- a/techniques/system/server-common/1.0/bootstrapRudderServer.cf
+++ b/techniques/system/server-common/1.0/bootstrapRudderServer.cf
@@ -1,3 +1,6 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+# SPDX-FileCopyrightText: 2021 Normation SAS
+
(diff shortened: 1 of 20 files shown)


techniques/system/server-common/1.0/bootstrapRudderServer.cf
```diff
@@ -0,0 +1,16 @@
+# Used only for bootstrapping Rudder server
+bundle agent bootstrap_server_rudder {
```
peckpeck, line 2: when is this called ?

ncharles: by good ol' rudder-init

ncharles: see https://github.com/Normation/rudder-packages/pull/2528

peckpeck (approved):

PR #1724, opened 2021-09-14
Fixes #19771: slow policy generation in Rudder 7.0

ncharles:
https://issues.rudder.io/issues/19771

With this, policy validation of server policies go from 2.2seconds to .4 seconds

### DIFF
--- a/techniques/system/common/1.0/cf-serverd.st
+++ b/techniques/system/common/1.0/cf-serverd.st
@@ -79,7 +79,7 @@ bundle common def
 
    # List here the IP masks that we grant access to on the server
 
-    policy_server::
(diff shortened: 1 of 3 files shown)


techniques/system/common/1.0/cf-serverd.st
```diff
@@ -79,7 +79,7 @@ bundle common def
 
    # List here the IP masks that we grant access to on the server
 
-    policy_server::
+    policy_server.(agent|server)::
```
ncharles, line 82: it doesn't need to be validated during cf-promises. it may help on large networks

amousset (approved):

PR #1726, opened 2021-10-08
Fixes #20096: Apache is not reloaded when the generated /var/rudder/lib/ssl/nodescerts.pem changes

Fdall:
https://issues.rudder.io/issues/20096

### DIFF
--- a/techniques/system/server-common/1.0/reloadRudderServices.st
+++ b/techniques/system/server-common/1.0/reloadRudderServices.st
@@ -11,14 +11,16 @@ bundle agent system_reload_rudder_services {
       "slapd_service_name"  string => "${rudder_slapd.service_name}";
 
       "jetty_prefix"          string => canonify("service_restart_${jetty_service_name}");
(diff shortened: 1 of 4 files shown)


techniques/system/server-common/1.0/reloadRudderServices.st
```diff
@@ -47,6 +49,8 @@ bundle agent system_reload_rudder_services {
     rudder_server_system_restart_jetty::
       "any" usebundle => service_restart("${jetty_service_name}");
     rudder_server_system_restart_apache::
+      "any" usebundle => service_restart("${apache_service_name}");
```
ncharles, line 52: with these changes, there never is any rudder_server_system_restart_apache defined - are you sure you haven't been too eager?

Fdall: It was never restarted, the "_restart_" class was only doing a reload.

ncharles (approved):

PR #1742, opened 2021-12-27
Fixes #20476: Don't insert partial runs as run logs - inventory detection doesn't work

amousset:
https://issues.rudder.io/issues/20476


ncharles (approved):

amousset:
OK, merging this PR

PR #1757, opened 2022-03-29
Fixes #20946: Increase default value of maxconnections

amousset:
https://issues.rudder.io/issues/20946


techniques/system/common/1.0/cf-serverd.st
```diff
@@ -135,7 +135,7 @@ body server control
         # force connections via tls1.2 or higher
         allowtlsversion => "1.2";
 
-        maxconnections    => "1000";
+        maxconnections    => "10000";
```
amousset, line 138: 1 connection = 1 thread so let's keep it reasonnable

ncharles (approved):

PR #1763, opened 2022-05-04
Fixes #20993: When updating allowed networks of a relay, rudder-cf-serverd service does not seem to be restarted

ncharles:
https://issues.rudder.io/issues/20993


techniques/system/server-common/1.0/reloadRudderServices.st
```diff
@@ -56,6 +60,8 @@ bundle agent system_reload_rudder_services {
       "any" usebundle => service_reload("${relayd_service_name}");
     rudder_server_system_restart_relayd::
       "any" usebundle => service_restart("${relayd_service_name}");
+    rudder_server_system_reload_apache::
```
amousset, line 63: why would apache reload restart cf-serverd?

ncharles: when the network changes

ncharles:
PR updated with a new commit

amousset (approved):

PR #1816, opened 2023-05-12
Fixes #22591: Variable from JSON file don't report in Rudder 7.2

Fdall:
https://issues.rudder.io/issues/22591

### DIFF
--- a/techniques/systemSettings/misc/variableFromJsonFile/3.0/metadata.xml
+++ b/techniques/systemSettings/misc/variableFromJsonFile/3.0/metadata.xml
@@ -27,7 +27,7 @@ For example a JSON containing `{ "datacenter": "paris" }` read into the variable
   </TRACKINGVARIABLE>
 
   <SECTIONS>
-    <SECTION name="Base file" component="true" componentKey="VARIABLE_FROM_JSON_FILE_JSON_FILE" />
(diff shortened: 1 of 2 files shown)


techniques/systemSettings/misc/variableFromJsonFile/3.0/metadata.xml
```diff
@@ -27,7 +27,7 @@ For example a JSON containing `{ "datacenter": "paris" }` read into the variable
   </TRACKINGVARIABLE>
 
   <SECTIONS>
-    <SECTION name="Base file" component="true" componentKey="VARIABLE_FROM_JSON_FILE_JSON_FILE" />
+    <SECTION name="Base file" component="true" componentKey="VARIABLE_FROM_JSON_FILE_BASE_FILE" />
```
ncharles, line 30: i'm dubious that it might work given that the componentKey is not related to the trackingVariable

ncharles:
with this PR, the situation is much worse, as the reporting get more invalid than withtout it
I've tested with 3 directives:
* one with 0 override
* one with 2 override
* one with 3 override
the one without override fails with current implementation, other works find
with this PR, all fail

Fdall:
This is an old PR and latest test tends to show that it does introduce new bugs.

PR #1840, opened 2024-03-12
Fixes #24412: Extend autoconditions to allow string values

amousset:
https://issues.rudder.io/issues/24412

Breaks compatibility with the corner cases described in the comments for array/object type, but I think it is for the better.

### DIFF
--- a/techniques/system/common/1.0/properties.cf
+++ b/techniques/system/common/1.0/properties.cf
@@ -88,7 +88,7 @@ bundle common properties
                                                                                                                                                  if => "is_string_${autocond_properties}_${${autocond_properties}_items}";
 
 # auto conditions cases:
-# - it the property is a boolean: define the condition if true (else do nothing)
+# - if the sub-property value is a boolean: define the condition if true (else do nothing)
 # - it the property is a string: define the condition with the value as suffix
 # - if the property has another type: do nothing
 # - if the property doesn't exist it doesn't fail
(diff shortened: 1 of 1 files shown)


techniques/system/common/1.0/properties.cf
```diff
@@ -68,18 +68,29 @@ bundle common properties
...
+    "${autocond_properties}_${${autocond_properties}_items}_${node.properties[${autocond_properties}][${${autocond_properties}_items}]}" expression => "true",
+                                                                                                                                                 if => "is_string_${autocond_properties}_${${autocond_properties}_items}";
 
-# auto conditions special cases:
-# - it the property is a string : it is ignored
-# - if the property is a json list such as [x,y...] : it is considered as a json object {"0": x, "1":y ... }
-# - if the property is a json object that contains sub levels : if there is a true value somewhere inside, it is considered true
-# - if the property is a json object that contains key/values with string values: the "true" string ins considered true, everything else is considered false
+# auto conditions cases:
+# - it the property is a boolean: define the condition if true (else do nothing)
```
Fdall, line 91:
```suggestion
# - if the sub-property value is a boolean: define the condition if true (else do nothing)
```

techniques/system/common/1.0/properties.cf
```diff
@@ -68,18 +68,29 @@ bundle common properties
...
+# - it the property is a string: define the condition with the value as suffix
```
Fdall, line 92:
```suggestion
# - if the sub-property is a string: define the condition with the value as suffix
```

techniques/system/common/1.0/properties.cf
```diff
@@ -68,18 +68,29 @@ bundle common properties
...
+# - if the property has another type: do nothing
```
Fdall, line 93:
```suggestion
# - if the property is not a Json type or its sub-properties are neither string nor boolean: do nothing
```

PR #1879, opened 2025-06-05
Fixes #26938: “SSH server” technique proposes deprecated “Use privilege separation” option

m-bouissou:
https://issues.rudder.io/issues/26938

Technique duplication before changes


m-bouissou:
PR updated with a new commit
Changes to the new version of the technique

m-bouissou:
Suggestions approved and commits squashed

PR #1893, opened 2025-07-11
Fixes #27267: Overwrite the /var/rudder/lib/ssl/policy_server.pem when it is a symlink

Fdall:
https://issues.rudder.io/issues/27267


fanf:
OK, merging this PR

